PR #13, opened 2026-06-06
fix(server): serve SPA index.html on deep-link/hard-reload instead of 404

cn0303:
## What does this PR do?

Hard-reloading or deep-linking any client-side route (e.g. `/recording`,
`/calibration`, `/teleoperation`) currently returns
`404 {"detail":"Not Found"}` and a blank page instead of the app.

The frontend is a single-page app: those routes live only in the browser's
router, not as files on disk. The `/` `StaticFiles` mount has no SPA fallback,
so any request for a path that isn't a real file 404s. The first load of `/`
and in-app navigation work (the browser handles routing client-side), so the
gap only surfaces on a hard reload or when a deep link is opened directly.

## How it's fixed

Add a small `SPAStaticFiles(StaticFiles)` subclass that, on a 404, falls back
to serving `index.html` so the SPA boots and its router renders the route.

The fallback only applies to **browser navigations** ie requests whose `Accept`
header actually wants `text/html`. API typos, XHR, and `curl` (which send
`application/json` or `*/*`) still receive a real JSON 404, so genuine API
errors are not masked. A dedicated `_accepts_html()` helper parses the header
and honors quality values, so an explicit `text/html;q=0` refusal is treated
as "no", not a substring match.

The mount stays last, after all API routes, so API endpoints continue to win.

## Before / After

| Request | Before | After |
| --- | --- | --- |
| Browser reload of `/recording` | `404` blank page | `200` app loads |
| `curl`/XHR for an unknown path | `404` JSON | `404` JSON (unchanged) |
| `GET /health`, `GET /` | `200` | `200` (unchanged) |

## Testing

- Added regression tests in `tests/test_server.py`: the HTML deep-link
  fallback, preservation of the JSON 404 for non-HTML clients, the
  `text/html;q=0` refusal path, and a parametrized table for `_accepts_html`.
- `pytest` and `pre-commit` (ruff, mypy, bandit, …) pass locally.
- Verified manually: ran `lelab`, hard-reloaded `/recording` and
  `/calibration` in the browser — pages load instead of the 404 blank screen.

### DIFF
--- a/lelab/server.py
+++ b/lelab/server.py
@@ -31,6 +31,10 @@ from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
 from fastapi.staticfiles import StaticFiles
 from pydantic import BaseModel
+from starlette.datastructures import Headers
+from starlette.exceptions import HTTPException as StarletteHTTPException
+from starlette.responses import Response
+from starlette.types import Scope
 
 from . import datasets as dataset_browser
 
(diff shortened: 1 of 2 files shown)


nicolas-rabault (approved):
Nice work on this 🙌 Clean, well-scoped fix for a real UX bug. 
Thanks for the thorough PR! 🚀

nicolas-rabault:
The pre-commit error is on my side, let me fix it.

cn0303:
Thanks @nicolas-rabault! Glad to see it merged and appreciate the feedback as well!

PR #14, opened 2026-06-07
feat: enable recording episodes into existent dataset

### LINKED ISSUE #11: Enable recording episodes into existent dataset
Currently it's only possible to record episodes to a newly created dataset. It would be great to have the option to record episodes to an existent dataset. 

The backend seems to allow resuming a dataset recording so only frontend changes should be required.

JeronimoMendes:
This PR: 
- extends the edit dataset page to support adding new episodes 
- shows sync status with hf hub. 
- deletes upload dataset page in favor of edit dataset. 
- matches styling of components to the job page.

## Demo 

shot 2026-06-07 at 02 11 20 

shot 2026-06-07 at 02 48 53@2x

## Disclaimer
AI tools such as Pi and Claude were used, specially on the frontend part of the PR, although I reviewed all of the code generated. 

Closes #11


nicolas-rabault (changes requested):
Very nice PR @JeronimoMendes 
A few simple modifications : 
1. Dead code

The resume feature is driven entirely from EditDataset.tsx, which renders RecordingSettingsFields inline and calls rec.startRecording directly. It never uses RecordingModal. Consequences:

**useRecording.ts:48** `openForResume(...)` is exported but never called anywhere.
The entire `isResuming` branch of **RecordingModal.tsx** is unreachable: the modal is only mounted in **Landing.tsx**, which always opens it via openForNew → resumeRepoId is always null there. So "Record More Episodes" / "Resume Recording" / the resume-repo display block / the "Append new episodes…" description can never render.
So the modal carries a resumeRepoId prop + branch and the hook carries an openForResume path that nothing exercises. Either wire the modal up for resume, or strip the resume branch from RecordingModal and drop openForResume. Given EditDataset already inlines the fields, the latter (delete) is the simpler, consistent choice.

2. Sync-status mtime comparison is fragile (medium)

`handle_dataset_sync_status` uses `local_root.stat().st_mtime` (top-level dataset dir) vs Hub lastModified. Two failure modes:

A directory's mtime only changes when a direct child entry is added/removed/renamed, not when files deep in data/chunk-*/ or videos/ are written, nor when an existing top-level file is rewritten in place. New episodes can land without the top-level mtime advancing, so needs_sync can be falsely False. The worst direction (UI says "in sync" when it isn't).
After a fresh download/clone from the Hub, local mtime = download time > hub_mtime, so it reports falsely needs_sync: True.
A more reliable signal would be a max mtime walk over the dataset tree, or reading the dataset's own episode count/revision metadata.

JeronimoMendes:
Thanks a lot for the quick review @nicolas-rabault .

1. Totally right about that, should be fixed now.
2. Thought about this and reached the conclusion that a simple file count might not be enough (I'm already thinking in the future with a possible episode deletion feature 😉, where a delete followed by a new episode would be considered "synced" with the hub). In fact, this is a far more complex issue than what this feature branch calls for because to truly be in sync with the Hub we need to consider pushes and pulls from the hub. To keep it simple, I've changed it so that we compute the local manifest (files + sizes) and match it against the one on the hub whilst still keeping the local dataset as the reference. Full bidirectional sync seems out of scope for this branch, let me know what you think!

frontend/src/hooks/useRecording.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    const cameraDict = cameras.reduce(
+      (acc, cam) => {
+        acc[cam.name] = {
+          type: cam.type,
+          camera_index: cam.camera_index,
+          width: cam.width,
+          height: cam.height,
+          fps: cam.fps,
```
nicolas-rabault, line 118: Since this adds appending, the resume path should take fps from the existing dataset rather than the inherited fps: 30. useEditDataset already fetches datasetInfo.fps, so thread that into the resume recordingConfig. Otherwise, resuming a dataset recorded at any other rate fails the compatibility check.

nicolas-rabault (changes requested):
Thanks,
I still have one minor change about Camera FPS.

nicolas-rabault (dismissed):
Thank a lot,
Could you fix the pre-commit? 
Then I will Merge this 👍.

JeronimoMendes:
Should be fixed, thanks a lot for taking the time to review the PR, Nicolas!

frontend/src/hooks/useRecording.ts
```diff
@@ -0,0 +1,181 @@
...
```
nicolas-rabault, line 118: The old `Landing.tsx` cameraDict forwarded the optional `fourcc` and `backend` fields, and the backend reads them (`_build_camera_configs` in `lelab/record.py` calls `camera_data.get("backend")` / `.get("fourcc")`). They're missing here, so UI overrides never reach the backend and silently fall back to defaults.

```suggestion
          type: cam.type,
          camera_index: cam.camera_index,
          width: cam.width,
          height: cam.height,
          fps: cam.fps,
          ...(cam.fourcc ? { fourcc: cam.fourcc } : {}),
          ...(cam.backend ? { backend: cam.backend } : {}),
```

Also add `fourcc?: string;` and `backend?: string;` to the `cameraDict` type annotation just below so TS keeps them in sync.

nicolas-rabault (changes requested):

frontend/src/hooks/useRecording.ts
nicolas-rabault, line 118: This is my 6 July comment, still open. `cameraDict` drops `fourcc` and `backend`.

It is a regression, not just an omission: `Landing.tsx` on `main` forwards both (lines 181-182), the backend reads both (`_build_camera_configs` calls `camera_data.get("backend")` and `.get("fourcc")`), and there is a test pinning that behaviour (`test_build_camera_configs_passes_fourcc_through`). Since this hook replaces the Landing code path, any per-camera pixel format or backend the user picks now silently falls back to the platform default on both the new-dataset and the resume flow.

```suggestion
          type: cam.type,
          camera_index: cam.camera_index,
          width: cam.width,
          height: cam.height,
          fps: cam.fps,
          ...(cam.fourcc ? { fourcc: cam.fourcc } : {}),
          ...(cam.backend ? { backend: cam.backend } : {}),
```

Add `fourcc?: string;` and `backend?: string;` to the `cameraDict` type annotation below so TypeScript keeps the two in sync.

lelab/record.py
```diff
@@ -518,6 +522,47 @@ def handle_delete_dataset(request: DatasetInfoRequest) -> dict[str, Any]:
...
+        for path in base.rglob("*"):
+            if path.is_file():
+                manifest[path.relative_to(root).as_posix()] = path.stat().st_size
+    return manifest
+
+
+def handle_dataset_sync_status(request: DatasetInfoRequest) -> dict[str, Any]:
+    """Detect unsynced local changes by diffing the local file manifest against the Hub's."""
+    repo_id = request.dataset_repo_id
+    local_root = Path(HF_LEROBOT_HOME).resolve() / repo_id
```
nicolas-rabault, line 543: `repo_id` reaches the filesystem with no containment check. `handle_delete_dataset`, directly above this function, resolves the target and rejects anything that escapes `HF_LEROBOT_HOME`, and this handler should do the same.

I confirmed it walks outside the cache: posting `{"dataset_repo_id": "../../../.claude/jobs/.../leaktarget"}` to `/dataset-sync-status` builds the manifest from that directory and only fails later, inside `api.dataset_info`. The response would carry the file count of an arbitrary directory tree. Read-only, but this server is also what the Space runs.

Same three lines as the delete handler:

```python
    root = Path(HF_LEROBOT_HOME).resolve()
    local_root = (root / repo_id).resolve()
    if local_root == root or root not in local_root.parents:
        return {"on_hub": False, "needs_sync": False, "local_files": 0, "hub_files": 0}
```

lelab/record.py
```diff
@@ -518,6 +522,47 @@ def handle_delete_dataset(request: DatasetInfoRequest) -> dict[str, Any]:
...
+
+    if not local_root.exists():
+        return {"on_hub": False, "needs_sync": False, "local_files": 0, "hub_files": 0}
+
+    local = _local_manifest(local_root)
+
+    api = shared_hf_api()
+    try:
+        info = api.dataset_info(repo_id, files_metadata=True)
+    except RepositoryNotFoundError:
```
nicolas-rabault, line 553: Only `RepositoryNotFoundError` is caught, so every other Hub failure escapes as an untyped 500. I reproduced it two ways: with the Hub unreachable the call raises `httpx.ConnectError`, and a `repo_id` that fails `validate_repo_id` raises `HFValidationError`. Both give a bare `Internal Server Error`.

The frontend consequence is silent. `useEditDataset.refreshSyncStatus` logs to the console and leaves `syncStatus` at `null`, so `isSynced` and `needsPush` are both false and `EditDataset` renders the sync strip as an empty grey box with no Publish button. Recording next to the robot with no network is a normal way to run this app, and in that state the page loses its upload path with no explanation.

Catch the connection and validation failures too and return a third state (`"unknown"`, or `on_hub: null`) that the panel can render as "could not reach the Hub", keeping Publish available.

nicolas-rabault (changes requested):
Thanks for this, and sorry for the long wait. Let’s keep all three features together in this PR. Please address the remaining fixes detailed inline before we merge.

PR #21, opened 2026-06-08
fix(calibration): don't open cameras until the user turns them on

cn0303:
## What does this PR do?

Opening or reloading the **Calibration** page used to switch on the
webcam and any other connected camera without the user doing anything.
This PR gates the camera section behind an explicit **On/Off toggle** (off by
default), so cameras are only touched once the user opts in.

## The bug

The "Attached cameras" section is rendered inline on the calibration page and
mounts on load. On mount, `useAvailableCameras` fires a 
`getUserMedia({ video: true })` permission probe to unlock device labels for 
enumeration. Because `{ video: true }` lets the browser pick the device, the 
probe grabs whatever the OS treats as the **default** camera independent of the 
robot's config. So:

- just visiting/reloading the page turns a camera on, and
- a camera can light up **even after being removed** from the configuration,
  because the probe doesn't care what's configured.

## The fix

- Add an **On/Off `Switch`** to the section header, **off by default**.
- While off, `CameraConfiguration` is **not mounted**, so nothing calls
  `getUserMedia` and no camera is touched. The panel instead explains that
  turning cameras on will scan connected devices and trigger the browser
  permission prompt ("Nothing is recorded").
- Turning it on mounts the section (scan + previews); turning it off unmounts
  it and releases every stream via the existing hook cleanups.
- Adds a **Rescan** button next to the camera list (a small refresh control,
  spins while scanning) so a newly plugged-in USB camera can be picked up on
  demand instead of waiting for the `devicechange` auto-refresh. Lives in the
  shared `CameraConfiguration`, so the recording modal gets it too.

## Before / after

| Action | Before | After |
| --- | --- | --- |
| Open/reload calibration page | camera(s) switch on unprompted | nothing turns on |
| Camera removed from config | can still be grabbed by the probe | not touched while off |
| Turn the toggle on | n/a | scans + previews, asks permission |
| Plug in a USB camera | wait for auto-refresh | click Rescan |

### Before UI
image

### After UI
image
image

## Testing

Frontend only. Verified manually: loaded the calibration page and confirmed no
camera activates until the toggle is switched on; toggling off releases the
streams; Rescan refreshes the list after a hotplug. `npm run build` passes, and
the changed files pass ESLint directly:

    npx eslint src/pages/Calibration.tsx src/components/recording/CameraConfiguration.tsx

### DIFF
--- a/frontend/src/components/recording/CameraConfiguration.tsx
+++ b/frontend/src/components/recording/CameraConfiguration.tsx
@@ -10,7 +10,7 @@ import {
 } from "@/components/ui/select";
 import { Input } from "@/components/ui/input";
 import { NumberInput } from "@/components/ui/number-input";
-import { Camera, Plus, X, VideoOff } from "lucide-react";
+import { Camera, Plus, X, VideoOff, RefreshCw } from "lucide-react";
 import { useToast } from "@/hooks/use-toast";
 import { useAvailableCameras } from "@/hooks/useAvailableCameras";
 import { useCameraStream } from "@/hooks/useCameraStream";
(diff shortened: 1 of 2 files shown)


nicolas-rabault (approved):
Clean, well-scoped fix and the right diagnosis.

Verified on macOS: nothing activates on load, toggle on scans + previews, toggle off releases, Rescan picks up a hotplug.

PR #25, opened 2026-06-08
fix(teleop): report connection failures and stop cleanly on exit

### LINKED ISSUE #24: Bug: Teleoperation failures are silent and sessions leak on exit
### Ticket Type
🐛 Bug Report

### Summary

Two related problems make teleoperation currently unreliable and confusing:
1. Starting teleoperation always "succeeds", even with no arm connected : the
   user is taken to a blank teleop screen with no explanation.
2. Leaving the teleop page doesn't stop the session : so the next start fails with 
   "Teleoperation is already active".

### 1. Failures are silent → blank screen

`handle_start_teleoperation` (`lelab/teleoperate.py`) spawns a background worker
thread that does the actual `robot.bus.connect()` / `teleop_device.bus.connect()`,
and returns `{"success": True}` **immediately** before the worker has tried to
connect. If the connect then fails (arm unplugged, powered off, port busy), the
exception is caught and only logged; it never reaches the caller. There's also a
cleanup gap: if the follower connects but the leader then fails, the follower is
never disconnected, so its serial port stays held.

Apart from this, the frontend (`RobotConfigManager.tsx`) navigates on `if (res.ok)`
 the HTTP status while the endpoint returns HTTP 200 with `{"success": false}`
for logical failures. So the UI navigates regardless of `data.success`.

Result: click → HTTP 200 → navigate → the worker fails to connect in the
background → blank teleop screen (the WebSocket connects, so it can even show a
green "Live Robot Data" indicator while the robot never moves).

### 2. Sessions leak on exit

The teleop page only stops teleoperation in the on-screen back button's handler.
Leaving any other way typing a URL, reloading, closing the tab, browser back
never calls stop, so `teleoperation_active` stays `True` and the next start
returns "Teleoperation is already active".

### Proposed fix

- Connect synchronously in the start handler so a failure is returned to the
  caller (`success: false`) with a message naming which arm failed
  (leader/follower) and asking the user to retry. Clean up any partially
  connected device. Only the streaming loop stays in the background thread.
- Frontend gates on `data.success`, not just `res.ok`.
- Stop on every exit from the teleop page: React unmount for in-app
  navigation, and a `pagehide` keepalive request for browser-level leaves
  (URL/reload/close), with a one-time confirmation toast wherever the user lands.

The same fire-and-forget + `res.ok`-only pattern likely affects the recording and
inference start flows; scoping this fix to teleoperation and noting the others as
follow-ups.

### Broader follow-up (separate): connection status on the landing page

The robot tile shows **"Ready"** whenever the robot is *configured/calibrated*
(`is_clean`), regardless of whether the arm is actually connected so there's no
up-front signal either. I think a real "connected" indicator would be useful, but it
needs some design modifications to make a non-intrusive way to detect connection 
(without opening the port on every poll) and a robot-type-agnostic status model so 
it scales beyond the SO-101 leader/follower setup (e.g. Reachy, AmazingHand, 
which may have neither a leader/follower pair nor a serial port). Raising it here 
as the broader goal; will bring it up in detail later on.

### Note

Opening a PR based on the above - if there are any modifications or alterations we can 
come back to this and update the approach.

cn0303:
Fixes #24 

## What does this PR do?

Makes teleoperation honest about whether it actually started, and stops the
session when the user leaves the page. Two user-visible bugs:

1. Starting teleoperation always succeeded and navigated to the teleop screen,
   even when no arm was connected.
2. Leaving the teleop page by anything other than the in-app back button left the
   session running, so the next start failed with "Teleoperation is already
   active".

## The bug

- Backend reported success before connecting - `handle_start_teleoperation`
  spawned a worker thread that did the real `bus.connect()`, and returned
  `{"success": true}` *immediately*. A failed connect was caught and only logged.
  (Plus a cleanup gap: a follower that connected before the leader failed was
  never disconnected, holding its port.)
- Frontend only checked the HTTP status - `handleTeleop` navigated on
  `if (res.ok)`, but the endpoint returns HTTP 200 with `{"success": false}` for
  logical failures so it navigated regardless of `data.success`.
- Teleop session leaked on exit - Stop was only wired to the on-screen back
  button, so URL navigation / reload / tab close never stopped it.

## The fix

Backend (`lelab/teleoperate.py`)
- Connect to both arms synchronously before returning, so a failure is
  reported to the caller as `success: false` with a message that names the arm
  that failed (leader vs follower) and how to retry. Only the streaming loop
  runs in the background thread.
- Add `_safe_disconnect()` and use it to release a partially-connected device on
  failure (and in the worker's cleanup), so a serial port is never left held.

Frontend
- Gate teleop navigation on `data.success`, not just `res.ok`; failures now show
  an error toast and keep the user on the landing page.
- Stop teleoperation on every exit from the teleop page:
  - in-app navigation → React unmount cleanup (stops + toasts immediately);
  - browser-level leave (URL change, reload, tab close) → a `pagehide` keepalive
    request that survives the unload, plus a `sessionStorage` flag that a small
    top-level `TeleopStopNotice` reads on the next load to confirm the clean
    disconnect wherever the user lands.

## Before / after

| Action | Before | After |
| --- | --- | --- |
| Start with arm unplugged | "success", teleop screen with robot | error toast naming the missing arm; stays on landing |
| Start with arm connected | works | works (now connects before responding) |
| Leave teleop via URL / reload / tab close | session keeps running → next start "already active" | session stops cleanly; confirmation shown |
| Leave teleop via back button | stops | stops (+ toast) |

### Image of failed connection when an arm is not connected -
image

## Testing

- Adds two regression tests (`tests/test_teleoperate.py`): one mocks a device
  whose connect raises and asserts the start handler returns `success: false`
  with a message naming the failed arm and resets state so a retry isn't
  blocked; the other covers the partial-connect path (follower connects, leader
  fails) and asserts the already-connected follower is disconnected so its port
  is released.
- `pytest` and `pre-commit` (ruff, mypy, bandit, …) pass; `npm run build` passes
  and the changed files pass ESLint:

      npx eslint src/pages/Teleoperation.tsx src/components/TeleopStopNotice.tsx \
        src/App.tsx src/components/landing/RobotConfigManager.tsx

- Verified manually on Windows with an SO-101 leader/follower: unplugged → clear
  error; connected → teleop works; every exit path (back button, browser back,
  URL navigation, reload) stops the session and shows the confirmation, and a
  subsequent start works without "already active".

## Out of scope for this bug
The robot tile still shows "Ready" based on configuration/calibration, not live
connection. A real "connected" indicator needs a non-intrusive connection check
and a robot-type-agnostic status model (to scale beyond SO-101 leader/follower);
noted in the issue as the broader goal.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -15,6 +15,7 @@ import Upload from "@/pages/Upload";
 
 import NotFound from "@/pages/NotFound";
 import SingleTabGuard from "@/components/SingleTabGuard";
+import TeleopStopNotice from "@/components/TeleopStopNotice";
 import { TooltipProvider } from "@radix-ui/react-tooltip";
 import { ApiProvider } from "./contexts/ApiContext";
 import { HfAuthProvider } from "./contexts/HfAuthContext";
@@ -32,6 +33,7 @@ function App() {
                 <DragAndDropProvider>
                   <BrowserRouter>
                     <SingleTabGuard>
+                      <TeleopStopNotice />
                       <Routes>
                         <Route path="/" element={<Landing />} />
                         <Route path="/teleoperation" element={<Teleoperation />} />
--- a/frontend/src/components/TeleopStopNotice.tsx
+++ b/frontend/src/components/TeleopStopNotice.tsx
@@ -1,0 +1,36 @@
+import { useEffect } from "react";
+import { useToast } from "@/hooks/use-toast";
+
+const FLAG = "lelab:teleop-stopped";
+
+/**
+ * One-time confirmation that teleoperation was stopped during the previous
+ * page's unload (a browser navigation, reload, or tab close from the
+ * teleoperation page set a sessionStorage flag, since React cleanup can't run
+ * in those cases). On the next fresh load we surface a toast wherever the user
+ * landed, then clear the flag. In-app navigation away from teleop toasts
+ * directly and never sets the flag, so this never double-fires.
+ */
+const TeleopStopNotice = () => {
+  const { toast } = useToast();
+
+  useEffect(() => {
+    let stopped = false;
+    try {
+      stopped = sessionStorage.getItem(FLAG) === "1";
+      if (stopped) sessionStorage.removeItem(FLAG);
+    } catch {
+      /* sessionStorage unavailable — nothing to show */
+    }
+    if (stopped) {
+      toast({
+        title: "Teleoperation stopped",
+        description: "The arm was disconnected cleanly when you left the page.",
+      });
+    }
+  }, [toast]);
+
+  return null;
+};
+
+export default TeleopStopNotice;
--- a/frontend/src/components/landing/RobotConfigManager.tsx
+++ b/frontend/src/components/landing/RobotConfigManager.tsx
@@ -45,7 +45,10 @@ const RobotConfigManager: React.FC<RobotConfigManagerProps> = ({
         }),
       });
       const data = await res.json();
-      if (res.ok) {
+      // The backend returns HTTP 200 with `{ success: false }` for logical
+      // failures (arm not connected, already active), so gate on `data.success`
+      // — not just `res.ok` — or we'd navigate to an empty teleop screen.
+      if (res.ok && data.success) {
         toast({
           title: "Teleoperation Started",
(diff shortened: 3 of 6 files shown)


nicolas-rabault (approved):
Very nice improvement.
Thank you

cn0303:
Thank you!

PR #26, opened 2026-06-08
fix(cameras): return real device names on Windows and Linux

### LINKED ISSUE #12: Bug: camera preview cannot match browser devices on non-macOS because /available-cameras returns placeholder names
### Ticket Type
🐛 Bug Report

### Environment & System Info
```
LeLab v0.1.0 (uv tool install, HEAD 68d7b45) · opencv-python 4.13.0
LeRobot 0.5.2 · Python 3.12.13 · Windows 11 (10.0.26200) · PyTorch 2.10.0+cpu
Cameras: index 0 = wrist_cam, index 1 = built in webcam
```

### Description
On Windows, and likely Linux via the same non-macOS code path the camera preview never works: after adding a camera it shows **"No browser match"**, no feed, and the wrong cv2 index can be saved.

**Root cause:** the frontend maps a cv2 index → browser `deviceId` by **matching device names**, but the non-macOS backend returns placeholders. `GET /available-cameras` returns `{"index":0,"name":"Camera 0"}, {"index":1,"name":"Camera 1"}` (`server.py:936-948`), and `useAvailableCameras.ts:62-81` matches those against the browser's real labels → no match → `deviceId=""`. 
Only the macOS branch (`server.py:880-929`) returns real names, so non-macOS preview is broken by construction.

### Reproduction
1. With a built-in webcam + a USB camera, start `lelab`.
2. Camera Configuration → Add Camera → pick the USB cam → Add.
3. Preview shows "No browser match"; saved record has `device_id: ''` and `camera_index: 0` (wrong — USB cam is index 1).

Image

### Evidence
```
/available-cameras  -> [{"index":0,"name":"Camera 0"},{"index":1,"name":"Camera 1"}]
browser labels      -> ["USB2.0_CAM1 (05a3:9230)", "ASUS FHD webcam (2b7e:c711)"]
```
The placeholder names (`Camera 0`/`Camera 1`) match neither real browser label, so every entry falls through to `deviceId = ""` → "No browser match". Surfacing real device names on Windows/Linux (as the macOS branch already does) lets the existing matcher work.

### Checklist
- [x] Searched existing tickets (not a duplicate)
- [x] On latest `main` (HEAD 68d7b45)
- [x] Not env-specific → Reproduced on Windows; Linux uses the same placeholder-name branch, but I have not tested Linux locally.

### Possible Fix Direction
Return real device names on Windows (DirectShow FriendlyName, e.g. `pygrabber`, which keeps cv2 DSHOW index order) and Linux (`/sys/class/video4linux/video*/name`). This adds a dependency flagging for your input. 

**I would like to take this on as my first contribution**; happy to confirm the approach before opening a PR. 🙌

### LINKED ISSUE #16: "No browser match" for all cameras on Windows 11 despite cameras being detected
Image

Environment:
- Windows 11
- Installed via:
  uv tool install git+https://github.com/huggingface/leLab.git

Problem:
Browser detects cameras correctly:
- HD Camera
- Microsoft LifeCam

getUserMedia() works correctly.

OpenCV also detects cameras.

However LeLab shows "No browser match" for every camera.

GET /available-cameras returns:

{
  "status": "success",
  "cameras": [
    {
      "index": 0,
      "name": "Camera 0",
      "available": true
    },
    {
      "index": 2,
      "name": "Camera 2",
      "available": true
    }
  ]
}

When saving the robot configuration, the camera is stored as:

{
  "camera_index": 0,
  "device_id": ""
}

Log:

Saved robot record:
{
  "cameras": [
    {
      "camera_index": 0,
      "device_id": ""
    }
  ]
}

Expected behavior:
Camera preview should work and browser cameras should be matched correctly.

Actual behavior:
All cameras show "No browser match".

cn0303:
Fixes #12
Fixes #16

## What does this PR do?

`/available-cameras` returned placeholder names (`"Camera 0"`, `"Camera 1"`) on
Windows and Linux. The frontend matches a cv2 index to a browser `deviceId`
by name (the browser's `MediaDeviceInfo.label`), so a placeholder never
matched and the preview showed **"No browser match"** and the saved camera 
had an empty `device_id`. Only macOS worked, because it already returned real
AVFoundation names. Two independent reports: #12 and #16.

## The fix

Return each camera's real name, in the same order cv2 indexes it:

- Windows : read DirectShow FriendlyNames via **pygrabber** (new dependency), 
  which enumerates in the same order cv2's DSHOW backend indexes devices. 
  Windows recording is pinned to `DSHOW` so a `camera_index` opens the previewed
  device. pygrabber is imported under `try/except`, so if it's missing or COM
  init fails, enumeration falls back to the previous generic probe.
- Linux : read the v4l2 device name from `/sys/class/video4linux/video*/name`
  (stdlib only, no new dependency).
- macOS is unchanged.

With real names, the frontend's existing name-matching connects each camera to
its browser `deviceId` and the preview works.

## Dependency note

This adds **`pygrabber`** as a **Windows-only** dependency
(`pygrabber>=0.2 ; sys_platform == 'win32'`). Non-Windows installs get nothing
new, and the call site degrades gracefully if it's absent. I went ahead and
implemented it because two users (incl me) are blocked on Windows (#12, #16), a 
dep-free Windows path can't recover the cv2 index ordering reliably, and it's easy 
to adjust or revert. **Very happy to change the approach** (drop the dep, gate it
behind an extra, etc.) if you would prefer, just wanted to unblock with something
working and tested.

### Before the change
image

### After the change for SO-101 wrist cam 
image

## Testing

- Verified on Windows 11 with two cameras: `/available-cameras` now returns
  `"USB2.0_CAM1"` and `"ASUS FHD webcam"` (was `"Camera 0/1"`), the
  Calibration camera preview shows a live feed instead of "No browser match",
  and the saved camera has a real `device_id`.
- Added unit tests (mocked, so they run cross-platform): the Windows real-name
  path, the graceful fallback when pygrabber is unavailable, the Linux sysfs
  name lookup (present + missing), and that Windows recording is pinned to the
  DSHOW backend.
- `pytest tests/test_server.py tests/test_record.py`, `pre-commit`,
  `npm run build`, and the relevant ESLint checks pass locally.

### DIFF
--- a/lelab/record.py
+++ b/lelab/record.py
@@ -99,6 +99,10 @@ def create_record_config(request: RecordingRequest) -> RecordConfig:
         opencv_backend = Cv2Backends.AVFOUNDATION
     elif platform.system() == "Linux":
         opencv_backend = Cv2Backends.V4L2
+    elif platform.system() == "Windows":
+        # DirectShow, matching the order /available-cameras enumerates (via
+        # pygrabber) so a camera_index always opens the previewed device.
+        opencv_backend = Cv2Backends.DSHOW
     else:
         opencv_backend = Cv2Backends.ANY
 
--- a/lelab/server.py
+++ b/lelab/server.py
@@ -918,15 +918,77 @@ def _avfoundation_cameras_in_cv2_order() -> list[dict[str, Any]]:
         return []
 
 
+def _generic_cv2_cameras(backend) -> list[dict[str, Any]]:
+    """Last-resort enumeration: probe cv2 indices with placeholder names."""
+    import cv2
+
+    cameras: list[dict[str, Any]] = []
+    for i in range(10):
+        cap = cv2.VideoCapture(i, backend)
+        opened = cap.isOpened()
+        cap.release()
+        if opened:
+            cameras.append({"index": i, "name": f"Camera {i}", "available": True})
+    return cameras
+
+
+def _windows_cameras() -> list[dict[str, Any]]:
+    """Enumerate Windows cameras with their real DirectShow names.
+
+    pygrabber lists DirectShow video devices in the same order cv2's DSHOW
+    backend indexes them (which recording is pinned to), so the returned index
+    matches what ``cv2.VideoCapture(i, CAP_DSHOW)`` opens. The real names let the
+    frontend match each index to the browser's ``MediaDeviceInfo.label`` for the
+    live preview. Falls back to generic names if pygrabber is unavailable.
+    """
+    try:
+        from pygrabber.dshow_graph import FilterGraph
+
+        names = FilterGraph().get_input_devices()
+    except Exception as e:  # ImportError, or a COM/DirectShow failure
+        logger.warning("pygrabber unavailable; using generic camera names: %s", e)
+        import cv2
+
+        return _generic_cv2_cameras(cv2.CAP_DSHOW)
+    return [{"index": i, "name": name, "available": True} for i, name in enumerate(names)]
+
+
+def _v4l2_camera_name(index: int) -> str | None:
+    """Real camera name for /dev/video{index} from sysfs (Linux, no deps)."""
+    try:
+        with open(f"/sys/class/video4linux/video{index}/name", encoding="utf-8") as f:
+            return f.read().strip() or None
+    except OSError:
+        return None
+
+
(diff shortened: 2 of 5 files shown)


nicolas-rabault (approved):
It is still working on Mac, good fix.
Thank you.

Issue #30, opened 2026-06-09
LeLab falls back to CPU-only PyTorch despite CUDA-capable GPU


KirillZubarev:
LeLab falls back to CPU-only PyTorch despite CUDA-capable GPU

Environment
OS: Windows 10/11
Python: 3.12 (uv environment)
Installation: uv tool install lelab
GPU: NVIDIA GPU (visible via nvidia-smi)
CUDA: installed in system
PyTorch: CPU-only build (+cpu)
Problem

LeLab detects CPU-only execution even when a CUDA-capable GPU is available.

Startup logs:

Device 'cuda' is not available. Switching to 'cpu'
No accelerated backend detected. Using default cpu

System details:

GPU is detected via nvidia-smi
CUDA drivers are installed
Issue is caused by CPU-only PyTorch being installed in the environment
Expected behavior

LeLab should use CUDA when a compatible GPU is available.

Suggested improvement

When a mismatch is detected:

nvidia-smi detects NVIDIA GPU
torch.version.cuda is None OR torch.cuda.is_available() == False

LeLab should automatically install CUDA-enabled PyTorch instead of falling back to CPU.

PR #32, opened 2026-06-10
feat(system): warn when an NVIDIA GPU is present but PyTorch can't use CUDA

cn0303:
## What does this PR do?

Addresses #30. When an NVIDIA GPU is installed but the active PyTorch can't use
CUDA, LeRobot silently downgrades `--policy.device cuda` to CPU so the only signal
is one easily overlooked line mid-log:

```
Device 'cuda' is not available. Switching to 'cpu'.
No accelerated backend detected.
```

Training and inference then crawl on CPU with no actionable guidance. This PR
surfaces the mismatch proactively without reinstalling Pytorch:

- A prominent startup WARNING in the server terminal (where the user already
  sees LeRobot's log line) with the guidance of:
  ```
  ⚠️  NVIDIA GPU detected but PyTorch can't use CUDA (torch=2.10.0+cpu). Training
  and inference will run on CPU and may be much slower. To use the GPU, install a
  CUDA build of PyTorch. Pick your CUDA version at
  https://pytorch.org/get-started/locally/ (for example: pip install torch
  --index-url https://download.pytorch.org/whl/cu124), then restart LeLab.
  ```
- A new `GET /system/cuda-status` endpoint returning
  `{gpu_present, cuda_available, mismatch, torch_version, install_hint, docs_url}`
  so the frontend can show a banner (follow-up, see below).

## Why this happens (and why it mostly happens on Windows)

With a plain dependency install, PyTorch resolution is platform-dependent. The
uv PyTorch guide documents that PyPI hosts CPU-only wheels for Windows/macOS,
while Linux PyPI wheels are GPU-accelerated by default. CUDA-specific builds are
also published on PyTorch's own indexes (`download.pytorch.org/whl/cuXXX`) when
a project wants to opt into a particular CUDA variant.

LeRobot declares a `torch` dependency, but its accelerator docs tell users to
install a suitable PyTorch build for their backend ([lerobot accelerators](https://huggingface.co/docs/lerobot/torch_accelerators)). So on Windows,
installing LeLab can pull the CPU-only torch wheel transitively, exactly like the reporter's
`2.10.0+cpu`.

## Scope: "Addresses", not "Fixes"

The issue suggests auto-installing CUDA PyTorch. I deliberately didn't:

the correct CUDA wheel depends on the user's driver/CUDA version, the build is
large, and silently replacing a working `torch` at runtime can break the env (and
some users want CPU-only on purpose). So this PR fixes the part that's genuinely
LeLab's job i.e. turning a silent fallback into a louder, actionable one but
it doesn't make the GPU just work.

## How it works

- `lelab/utils/system.py`:
  - `_nvidia_gpu_present()` — dependency-free: `nvidia-smi` on PATH + `nvidia-smi
    -L` actually lists a GPU. Any failure → `False`.
  - `_torch_cuda()` — `(torch.cuda.is_available(), torch.__version__)`; missing/
    broken torch → `(False, None)`.
  - `detect_cuda_status()` / `warn_if_cuda_mismatch()` / `CudaStatus` model.
- `lelab/server.py`: `GET /system/cuda-status` + an `@app.on_event("startup")`
  hook calling `warn_if_cuda_mismatch()` (matches the file's existing
  `@app.on_event("shutdown")` style; happy to migrate both to `lifespan`).

## Notes

- No behavior change beyond the warning/endpoint; device selection is untouched.
- No-GPU machines (Macs, CPU boxes) are not nagged: CPU torch is expected there.

## Tests

`tests/test_utils_system.py` (+7): mismatch flagged (GPU + no-CUDA torch),
no-mismatch when CUDA available, no-mismatch without a GPU, `nvidia-smi` absent →
False, warn logs on mismatch / silent when OK, and the endpoint shape. `pytest` +
`ruff check` + `ruff format` + the full pre-commit (bandit, mypy) green; rest of
the suite unaffected.

## Possible follow-up

An opt-in "Install CUDA PyTorch" action driven by `/system/cuda-status`,
reusing LeLab's existing `InstallManager` (which already does `pip install
accelerate`/`wandb`) pointed at the CUDA index — showing the exact command before
running it, never silently. Plus a frontend banner so the warning is visible in
the UI, not just the terminal.

### DIFF
--- a/lelab/server.py
+++ b/lelab/server.py
@@ -99,12 +99,14 @@ from .utils.config import (
 )
 from .utils.hf_auth import cached_whoami, handle_hf_auth_status, handle_hf_login, shared_hf_api
 from .utils.system import (
+    handle_get_cuda_status,
     handle_get_training_extra,
     handle_get_wandb_extra,
     handle_install_training_extra,
     handle_install_training_extra_status,
     handle_install_wandb_extra,
     handle_install_wandb_extra_status,
+    warn_if_cuda_mismatch,
 )
 
 # Set up logging
@@ -691,6 +693,12 @@ def get_runners_hardware():
 # ============================================================================
 
 
+@app.get("/system/cuda-status")
+def get_cuda_status():
+    """Report whether an NVIDIA GPU is present but PyTorch is CPU-only (issue #30)."""
+    return handle_get_cuda_status()
+
+
 @app.get("/system/training-extra")
 def get_training_extra():
     """Return whether the LeRobot training extra (accelerate) is importable."""
@@ -1158,6 +1166,12 @@ def delete_robot(name: str):
     if delete_robot_record(name):
         return {"status": "success"}
     return JSONResponse(status_code=404, content={"status": "error", "message": "Robot not found"})
+
+
+@app.on_event("startup")
+def startup_event():
+    """One-time startup diagnostics surfaced in the server terminal."""
+    warn_if_cuda_mismatch()
 
 
 @app.on_event("shutdown")
--- a/lelab/utils/system.py
+++ b/lelab/utils/system.py
@@ -55,6 +55,15 @@ class ExtraStatus(BaseModel):
     install_hint: str
 
 
+class CudaStatus(BaseModel):
+    gpu_present: bool
+    cuda_available: bool
+    mismatch: bool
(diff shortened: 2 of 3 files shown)


nicolas-rabault (approved):
Perfect fix, thank you @cn0303 .

PR #42, opened 2026-06-19
feat(teleop): show live camera feeds beside the 3D viewer

cn0303:
### What
Adds an optional camera panel to the teleoperation page so you can watch your cameras live while
you drive the arm; handy for verifying framing and catching a camera that came loose.

- **Off by default.** Landing on the page never calls `getUserMedia`; the user opts in with a
  toggle; the same consent pattern as the calibration camera toggle (#21).
- **Mirrors the robot's configured cameras.** The panel shows one live feed per camera on the
  selected robot record (e.g. `wrist_cam`, `webcam`), stacked vertically with their names.
- **Strict mirror.** If the robot has no cameras configured, the panel shows nothing but a hint to
  add them during calibration — it never surfaces a device that wasn't deliberately configured.
- A configured-but-currently-absent camera still appears (name + "preview failed" placeholder) so
  it's obvious it's expected but not detected.

### Why this works during teleop
Teleoperation opens **no** cv2 cameras (`SO101FollowerConfig` has no `cameras`), so the browser is
free to stream them directly via `getUserMedia` while the arm runs. (Recording is different; there
cv2 owns the devices, so this browser-source approach is teleop-only by design.)

### Implementation notes
- `CameraFeed` is a small presentational component bound to a browser `deviceId`.
- `TeleopCameraPanel` reads the selected robot's cameras via `useRobots()` and renders a `CameraFeed`
  per configured camera. A retry button remounts the feeds for a fresh `getUserMedia` attempt that is
  handy if a camera was unplugged and reconnected. (No backend camera enumeration: the configured 
  cameras already carry the browser `device_id` we stream directly.)
- Mounted in the teleop `VisualizerPanel` via a new `rightSlot`.

### Testing
- Manual: configured cameras show stacked with correct names; zero-config shows the hint only;
  unplugged-but-configured shows the placeholder.
- `npx tsc --noEmit` clean; **ESLint clean on the changed files** (the repo's full `npm run lint` has
  pre-existing errors/warnings unrelated to this change); `npm run build` succeeds.
- Source-only PR — `frontend/dist` is left to CI's `build_frontend.yml` to rebuild on merge.

### DIFF
--- a/frontend/src/components/control/CameraFeed.tsx
+++ b/frontend/src/components/control/CameraFeed.tsx
@@ -1,0 +1,46 @@
+import React from "react";
+import { VideoOff } from "lucide-react";
+import { useCameraStream } from "@/hooks/useCameraStream";
+
+interface CameraFeedProps {
+  /** Browser deviceId to stream. Empty string renders the "no camera" state. */
+  deviceId: string;
+  /** Optional caption shown under the feed. */
+  label?: string;
+}
+
+/** Live browser-camera feed bound to a deviceId via getUserMedia. */
+const CameraFeed: React.FC<CameraFeedProps> = ({ deviceId, label }) => {
+  const { videoRef, hasError } = useCameraStream(deviceId, false);
+  const showVideo = deviceId && !hasError;
+
+  return (
+    <div className="bg-gray-900 rounded-lg border border-gray-700 overflow-hidden">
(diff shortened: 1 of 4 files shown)


nicolas-rabault (approved):
Very nice addition.

PR #45, opened 2026-06-25
Prompt to install policy extras (smolvla/pi0/diffusion) before training starts

nobullryder:
## Problem

Training a policy that pulls an optional LeRobot extra fails with a buried `ImportError` — but only **after** the training subprocess has already started, so the user just sees the run die in the logs:

```
ImportError: 'transformers' is required but not installed.
Install it with: pip install 'lerobot[smolvla]'
```

`smolvla`, `pi0`, and `pi0_fast` need `transformers`; `diffusion` needs `diffusers`.

## What this does

A pre-flight check on **Start Training**, built entirely on the existing install-extra machinery (`InstallManager`, `useInstallExtra`, `InstallProgress`, `WandbInstallDialog`) — no new infrastructure.

- `GET /system/policy-extra/{policy_type}` does a live `importlib.util.find_spec` probe and returns the install target. `POST .../install` + `GET .../install-status` reuse `InstallManager` to run `pip install lerobot[<extra>]` in the background.
- On Start Training the UI pre-flights the endpoint; if the extra is missing it opens a `PolicyExtraDialog` (a parameterized sibling of `WandbInstallDialog`) for a one-click install instead of starting a run that's doomed to crash.

| Policy | Probe | Install |
|---|---|---|
| `smolvla` | `transformers` | `lerobot[smolvla]` |
| `pi0`, `pi0_fast` | `transformers` | `lerobot[pi]` |
| `diffusion` | `diffusers` | `lerobot[diffusion]` |
| act, vqbet, tdmpc, sac, … | — | reported available (no-op) |

## Tests

`tests/test_utils_system.py` covers the policy→extra mapping, the `find_spec` availability logic, the core-policy no-op, the install no-op, and the `GET /system/policy-extra/{policy_type}` route.

## Notes

- After installing, **restart LeLab** so the new package is importable (same as the existing wandb/accelerate extras).
- Generated `frontend/dist` is included (built with `npm ci` against the committed lockfile, matching the `build_frontend` workflow) — please review the source files first.


nicolas-rabault (approved):
Very nice!
Thank you @nobullryder and welcome to the contributors ;)

PR #46, opened 2026-06-25
Reliable shutdown on Windows + `lelab --stop`

nobullryder:
## Problem

On Windows, stopping `lelab` (Ctrl-C / closing the terminal) often leaves the uvicorn and Vite child processes alive, which keeps port `:8000` — and any open camera handles — held. The next `lelab` then fails to bind, or a camera won't open because the previous run never released it.

## What this does

- **`lelab --stop`** — finds and terminates a running LeLab and its child process tree, freeing the port.
- **Clean process-tree teardown** on exit (psutil, with a Windows `taskkill /T` fallback) so the uvicorn/Vite children and their handles are actually released.
- **Pre-flight port checks** with an actionable message (*"…run `lelab --stop` to free it"*) instead of an opaque bind error, plus a readiness wait before opening the browser.

## Tests

`tests/test_scripts_lelab.py` covers the stop / teardown / port-check paths with mocked `psutil` + `subprocess` (no real processes spawned).

### DIFF
--- a/lelab/scripts/lelab.py
+++ b/lelab/scripts/lelab.py
@@ -36,6 +36,7 @@ import time
 import webbrowser
 from collections.abc import Sequence
 from pathlib import Path
+from typing import NoReturn
 
 import psutil
 import uvicorn
@@ -52,7 +53,7 @@ BACKEND_PORT = 8000
 FRONTEND_DEV_PORT = 8080
 
 
-def _fail(message: str) -> None:
+def _fail(message: str) -> NoReturn:
     logger.error(message)
     raise SystemExit(1)
 
@@ -148,7 +149,6 @@ def _require_command(command: str) -> str:
         f"`{command}` was not found on PATH. Install Node.js LTS from https://nodejs.org/, "
         "restart your terminal, then run LeLab again."
     )
-    raise AssertionError("unreachable")
 
 
 def _resolve_command(command: str) -> str | None:
@@ -163,7 +163,9 @@ def _resolve_command(command: str) -> str | None:
 def _ensure_frontend_path() -> None:
     if FRONTEND_PATH.exists():
         return
-    _fail(f"Frontend source not found at {FRONTEND_PATH}. Run LeLab from a complete checkout or reinstall it.")
+    _fail(
+        f"Frontend source not found at {FRONTEND_PATH}. Run LeLab from a complete checkout or reinstall it."
+    )
 
 
 def _ensure_frontend_dist() -> None:
@@ -243,7 +245,6 @@ def _start_process(
         return subprocess.Popen(resolved_command, **_child_process_kwargs(cwd, env))
     except FileNotFoundError as exc:
         _fail(f"Could not start {name}: {exc}. Check that `{command[0]}` is installed and on PATH.")
-    raise AssertionError("unreachable")
 
 
 def _terminate_tree(pid: int, timeout: int = 5) -> None:
@@ -308,10 +309,11 @@ def _open_browser_when_ready(port: int, no_open: bool) -> None:
         time.sleep(0.5)
 
 
-def _install_signal_handlers(processes: Sequence[tuple[str, subprocess.Popen]]) -> None:
+def _install_signal_handlers() -> None:
+    # Teardown of the child processes is owned by _run_dev's
+    # `except BaseException` — the handler only needs to unwind.
     def shutdown(_signum, _frame) -> None:
         logger.info("Shutting down LeLab...")
-        _shutdown_processes(processes)
         raise SystemExit(0)
 
     signal.signal(signal.SIGINT, shutdown)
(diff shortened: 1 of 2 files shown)


lelab/scripts/lelab.py
```diff
@@ -15,185 +15,462 @@
...
+        port=BACKEND_PORT,
+        log_level="info",
+        reload=False,
+    )
+    server = uvicorn.Server(config)
+    # On Windows, uvicorn's graceful shutdown frequently hangs on Ctrl+C (the
+    # asyncio Proactor loop doesn't wind down cleanly), leaving the terminal
+    # stuck. Take over signal handling: stop hard and reap any child
+    # subprocesses (training/recording/inference) so the prompt always returns.
+    server.install_signal_handlers = lambda: None
```
nicolas-rabault: The hard-exit path is gated by intent (the comment is about the Windows Proactor loop hanging) but not by code — `install_signal_handlers = lambda: None` and the `os._exit(0)` below run on every platform. On macOS/Linux this drops the graceful shutdown the old prod code explicitly opted into (`timeout_graceful_shutdown=2`) and skips the `@app.on_event("shutdown")` cleanup. Can we gate this to Windows (e.g. `if os.name == "nt":` around the override + hard exit) and keep uvicorn's native graceful handler elsewhere?

lelab/scripts/lelab.py
```diff
@@ -15,185 +15,462 @@
...
-            shutdown(None, None)
-        if frontend_process.poll() is not None:
-            logger.error("❌ Frontend died")
-            shutdown(None, None)
+        for name, process in processes:
+            returncode = process.poll()
+            if returncode is not None:
+                logger.error("%s stopped with exit code %s.", name, returncode)
+                _shutdown_processes(processes)
+                raise SystemExit(returncode or 1)
```
nicolas-rabault: `returncode or 1` turns a clean child exit (`returncode == 0`) into exit code `1`. Suggest `raise SystemExit(returncode if returncode is not None else 1)`.

lelab/scripts/lelab.py
```diff
@@ -15,185 +15,462 @@
...
+FRONTEND_NODE_MODULES = FRONTEND_PATH / "node_modules"
+HOST = "127.0.0.1"
 BACKEND_PORT = 8000
 FRONTEND_DEV_PORT = 8080
 
 
-def _wait_for_port(port: int, timeout: int = 30) -> bool:
-    for _ in range(timeout):
-        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+def _fail(message: str) -> None:
```
nicolas-rabault: If you annotate this `-> typing.NoReturn`, the type-checker knows control never returns past `_fail`, which lets you delete the unreachable `raise AssertionError("unreachable")` lines at 151 and 246 (they're only there because the return type currently looks like `None`).

lelab/scripts/lelab.py
```diff
@@ -15,185 +15,462 @@
...
+        if _is_port_open(port):
+            _open_browser_url(f"http://localhost:{port}/", no_open=False)
+            return
+        time.sleep(0.5)
+
+
+def _install_signal_handlers(processes: Sequence[tuple[str, subprocess.Popen]]) -> None:
+    def shutdown(_signum, _frame) -> None:
+        logger.info("Shutting down LeLab...")
+        _shutdown_processes(processes)
```
nicolas-rabault: Minor: this runs teardown, then the `raise SystemExit(0)` is caught by the `except BaseException` at line 430, which calls `_shutdown_processes` again. It's idempotent so harmless, but the handler could just `raise SystemExit(0)` and let the `except` own teardown to avoid the double pass.

nicolas-rabault (changes requested):
Thank you @nobullryder for this contribution.
Here are some changes requests that could improve it.

lelab/scripts/lelab.py
nobullryder: Done — the `install_signal_handlers` override and the hard-exit handler are now inside `if os.name == "nt":`. On macOS/Linux uvicorn keeps its native handlers, and `timeout_graceful_shutdown=2` is restored on the config so prod drains gracefully as before.

lelab/scripts/lelab.py
nobullryder: Done — since this is inside the `returncode is not None` branch it's now simply `raise SystemExit(returncode)`, so a clean child exit propagates as 0.

lelab/scripts/lelab.py
nobullryder: Done — `_fail` is annotated `-> NoReturn` and both `raise AssertionError("unreachable")` lines are gone.

lelab/scripts/lelab.py
nobullryder: Done — the handler now just raises `SystemExit(0)` and the `except BaseException` in `_run_dev` owns teardown. Applied the same to `_monitor_processes`, which had the identical double pass.

nicolas-rabault (approved):
Very nice, thank you 🤗

PR #47, opened 2026-06-25
Force-release camera/serial resources when a device disconnect fails

nobullryder:
## Problem

When LeRobot's `device.disconnect()` throws (a flaky USB camera or serial port), the underlying handle can stay open — which keeps the camera busy and blocks the next teleop / record / calibrate run from acquiring it.

## What this does

`safe_disconnect_device()` attempts the normal disconnect, and on failure best-effort **force-closes** the serial/camera resources so the device is actually released. It's wired into teleoperation's teardown here; calibration and recording can adopt the same helper in follow-ups.

## Tests

`tests/test_devices.py` covers the success path and the disconnect-fails-then-force-release path.


nicolas-rabault (approved):
Nicely done, thank you @nobullryder

PR #48, opened 2026-06-25
Surface the real inference error; don't report a gripper-overload shutdown as failed

nobullryder:
## Problem

When a rollout (inference) exits non-zero, the UI just said *"failed — check logs"*, which (a) buries the real error in a log file inside the HF cache, and (b) calls a **working** run a failure when only shutdown tripped — e.g. disabling torque on a gripper that's still holding an object trips an overload during cleanup.

## What this does (all in `rollout.py`)

- **`_extract_error_from_log`** — pulls the actual exception out of the rollout log so the UI can show it directly.
- **`_friendly_hint`** — a plain-language, actionable headline for the common SO-101 failures (gripper overload, unresponsive motor id 6, can't-connect, camera-too-slow, unsupported resolution, busy serial port).
- **`_classify_outcome`** — a non-zero exit *after* the rollout main loop started, where the error is a torque-disable/overload on shutdown, is reported as `ran_with_warning` rather than `failed`. The skill ran; only cleanup complained.

## Tests

`tests/test_rollout.py` adds coverage for the outcome classification, the hint mapping, and the log-error extraction.

### DIFF
--- a/lelab/rollout.py
+++ b/lelab/rollout.py
@@ -164,11 +164,14 @@ def _extract_error_from_log(log_path: str | None) -> str | None:
     if not log_path:
         return None
     try:
-        with open(log_path, encoding="utf-8", errors="replace") as fh:
-            lines = fh.readlines()
+        # Only the tail matters; avoid materializing a multi-MB verbose log.
+        with open(log_path, "rb") as fh:
+            fh.seek(0, os.SEEK_END)
+            fh.seek(max(0, fh.tell() - 64 * 1024))
+            data = fh.read()
     except OSError:
         return None
-    tail = [ln.rstrip("\n") for ln in lines[-50:]]
+    tail = data.decode("utf-8", errors="replace").splitlines()[-50:]
     # Prefer the last exception line + everything after it (the message body).
     exc_idx = next((i for i in range(len(tail) - 1, -1, -1) if _EXC_LINE_RE.match(tail[i])), None)
     if exc_idx is not None:
@@ -212,8 +215,9 @@ def _friendly_hint(error_text: str | None) -> str | None:
 
 
 # Errors that mean the policy actually ran and only shutdown/cleanup tripped —
-# e.g. disabling torque on a gripper still holding an object.
-_CLEANUP_MARKERS = ("overload", "torque_enable", "disconnect", "not connected")
+# e.g. disabling torque on a gripper still holding an object. Connection-loss
+# errors are deliberately excluded: a mid-run disconnect is a real failure.
+_CLEANUP_MARKERS = ("overload", "torque_enable")
 
 
 def _classify_outcome(rc: int | None, rollout_started: bool, error_text: str | None) -> str:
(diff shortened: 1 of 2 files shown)


lelab/rollout.py
```diff
@@ -153,6 +153,84 @@ def _format_cameras_arg(cameras: dict[str, dict[str, Any]]) -> str:
...
+    if "failed to set capture_" in low or "actual_width" in low or "actual_height" in low:
+        return "A camera doesn't support the configured resolution — open camera settings and click Auto."
+    if "permission" in low and ("port" in low or "com" in low):
+        return "Couldn't open the serial port — close anything else using it, or run `lelab --stop`."
+    return None
+
+
+# Errors that mean the policy actually ran and only shutdown/cleanup tripped —
+# e.g. disabling torque on a gripper still holding an object.
+_CLEANUP_MARKERS = ("overload", "torque_enable", "disconnect", "not connected")
```
nicolas-rabault: `disconnect` / `not connected` are broader than the "torque-disable/overload on shutdown" case this PR targets. Because `_classify_outcome` only checks that the rollout main loop *started* (not that it finished), a genuine mid-run connection loss — e.g. the USB cable gets bumped while the policy is running — raises `DeviceNotConnectedError` ("not connected"), exits non-zero, and gets reported as `ran_with_warning` even though the task crashed partway through.

The overload case is already covered by `overload` / `torque_enable`. Suggest dropping the two connection markers here, or gating the warning on the run having reached ~its requested duration rather than on the error string.

lelab/rollout.py
```diff
@@ -153,6 +153,84 @@ def _format_cameras_arg(cameras: dict[str, dict[str, Any]]) -> str:
...
+
+
+def _extract_error_from_log(log_path: str | None) -> str | None:
+    """Pull the meaningful error out of a failed rollout's log so the UI can
+    show it directly instead of telling the user to open a file in the cache."""
+    if not log_path:
+        return None
+    try:
+        with open(log_path, encoding="utf-8", errors="replace") as fh:
+            lines = fh.readlines()
```
nicolas-rabault: Nit (non-blocking): `readlines()` loads the whole log into memory just to slice `lines[-50:]`. Rollout logs are line-buffered over a 60s+ verbose run and can reach a few MB. Fine as-is given it only runs once on exit, but a tail read would avoid materializing the full file.

nicolas-rabault (changes requested):
thank you,
Here are some simple modification I would like.

lelab/rollout.py
nobullryder: Done — dropped `disconnect`/`not connected` from `_CLEANUP_MARKERS` so a mid-run connection loss is reported as a real failure; the overload case stays covered by `overload`/`torque_enable`. Added a regression test for the mid-run `DeviceNotConnectedError` case.

lelab/rollout.py
nobullryder: Done — now reads only the last 64KB of the log (binary seek from the end) instead of materializing the whole file.

nicolas-rabault (approved):
Thank you for this contribution 🤗

Issue #51, opened 2026-06-29
unrecognized arguments during training: --eval_freq 0


Tibe31:
Hello,

I'm trying to train a new model on cloud but I'm getting this error: 

`lerobot_train.py: error: unrecognized arguments: --eval_freq 0
[wrapper] trained exited with rc=2`

I'm using LeLab interface.

I opened a PR with a minimal fix here: #52

PR #52, opened 2026-06-29
Fix LeRobot env eval frequency flag

Tibe31:
This PR updates LeLab's training command builder to use the current LeRobot CLI flag for environment evaluation frequency.
Fixes Issue #51 

LeRobot renamed `--eval_freq` to `--env_eval_freq` in huggingface/lerobot#3824. LeLab still emits the old flag, causing cloud training jobs to fail before training starts with:

```text
lerobot_train.py: error: unrecognized arguments: --eval_freq 0

### DIFF
--- a/lelab/train.py
+++ b/lelab/train.py
@@ -135,7 +135,7 @@ def build_training_command(
     # Logging / checkpointing
     cmd.extend(["--log_freq", str(request.log_freq)])
     cmd.extend(["--save_freq", str(request.save_freq)])
-    cmd.extend(["--eval_freq", str(request.eval_freq)])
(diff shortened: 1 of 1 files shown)


nicolas-rabault:
Thank you @Tibe31,
The flag rename is correct for the latest lerobot version but breaks the currently pinned version on main, and the fix is already incorporated in the upcoming port/lerobot-0.6.0 branch. So it'll ship with the 0.6.0 release.

Tibe31:
Thanks for clarifying @nicolas-rabault. Good to know it’ll ship with the 0.6.0 release.

PR #60, opened 2026-07-02
Feat/gripper force feedback

jedld:
Adds optional **gripper-only force feedback** for SO-101 teleoperation: when the follower grasps an object, the leader gripper resists squeezing past the follower's achieved position (virtual wall), with no feedback while releasing.
- Exposes a **Gripper force feedback** checkbox on the landing page robot tile and optional `gripper_force_feedback` / `gripper_force_feedback_gain` fields on `POST /move-arm`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -36,7 +36,7 @@ Run the Python tests with `pytest` (config in [pyproject.toml](pyproject.toml); 
 [server.py](lelab/server.py) is a thin FastAPI router. Each feature lives in its own module that owns its global state and exposes `handle_*` functions plus a Pydantic request model:
 
 - [record.py](lelab/record.py) — dataset recording (wraps `lerobot.record.record`); patches `lerobot.common.utils.control_utils` keyboard listener so frontend buttons replace arrow-key controls.
-- [teleoperate.py](lelab/teleoperate.py) — leader→follower teleoperation (wraps `lerobot.teleoperate`).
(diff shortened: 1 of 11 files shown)


nicolas-rabault:
Thank you so much for this.
It's a genuinely nice piece of work. The grasp detection, virtual-wall logic, release hysteresis, and torque shaping are well thought out, the docs are excellent, and it's clear you tested carefully on real hardware. 🙏

I want to be upfront about scope so this effort lands where it'll have the most impact. LeLab is deliberately a thin web wrapper whose only job is to make LeRobot easier to use; it shouldn't hold robot-control logic. The core of this PR (reading follower `Present_Load`/`Present_Position`, managing leader torque and `Torque_Limit`, and the whole feedback control policy) is exactly that kind of logic. It belongs in **[LeRobot](https://github.com/huggingface/lerobot)**, where every consumer benefits, not just LeLab users.

The good news is LeRobot already has the plumbing for this: `Teleoperator.send_feedback` / `feedback_features` exist, and the teleoperate loop already calls `send_feedback` for the Unitree G1 (`if robot.name == "unitree_g1"`). A grasp-aware feedback policy for the SO leader would slot right into that same path. Once it's upstream, LeLab's side becomes trivial, just a checkbox that toggles the existing LeRobot flag.

So I'm going to close this here, but I'd really encourage you to open it against LeRobot. I think it's a great contribution and I'd be happy to help however I can.

One more thing: the teleoperation-robustness parts of this PR (tolerating transient serial errors, `stop_reason` plumbing, the status poller, and the bfcache `pagehide` fix) *are* legitimately LeLab-layer and don't depend on the feedback logic. If you'd like, those would be very welcome as a separate, focused PR here.

Thanks again for the thoughtful contribution! 🤗

PR #67, opened 2026-07-13
refactor(record): show configured cameras read-only in the recording dialog

cn0303:
## What does this PR do?

The Configure Recording dialog embedded the full camera-configuration UI (Add
Camera, the available-cameras dropdown, plus a remove button and per-camera edit
fields on each preview). That duplicated the camera setup already done on the
Calibration page and stored on the robot record, and let a user add a camera in
one place that the robot config in the other place did not know about.

This makes the dialog show the robot's configured cameras read-only: the live
previews stay, but the add, remove, and edit controls are gone, with a note
pointing to the Calibration page to change cameras. Camera setup now lives in one
place.

## How it works

- `CameraConfiguration` gains a `readOnly` prop (default false). When set, it
  hides the Add Camera section, and each preview drops its remove button and its
  editable Configuration panel, showing the camera's resolution as static text
  instead. The live preview itself is unchanged.
- `RecordingModal` renders `CameraConfiguration` with `readOnly` and adds the
  note.
- The Calibration page uses `CameraConfiguration` without `readOnly`, so its full
  editable setup is unchanged.

Recording behaviour is unchanged: the dialog was already seeded with the cameras
from the selected robot's record, and the recording request still sends exactly
those. The stream-release step before recording starts is also unchanged, so the
browser previews are released before cv2 opens the cameras.

## Testing

Verified on Windows 11 with a configured SO-101 (wrist_cam + webcam): the dialog
shows both live previews read-only with no add/remove/edit controls, the note is
present, and recording still starts and captures both cameras. The Calibration
page still adds and edits cameras normally.

ESLint, `tsc --noEmit`, and `npm run build` pass. Source only; `frontend/dist` is
left to CI to rebuild on merge.

### DIFF
--- a/frontend/src/components/landing/RecordingModal.tsx
+++ b/frontend/src/components/landing/RecordingModal.tsx
@@ -234,12 +234,17 @@ const RecordingModal: React.FC<RecordingModalProps> = ({
               </div>
             </div>
 
-            <div className="space-y-4">
+            <div className="space-y-3">
               <CameraConfiguration
                 cameras={cameras}
                 onCamerasChange={setCameras}
                 releaseStreamsRef={releaseStreamsRef}
+                readOnly
               />
+              <p className="text-xs text-gray-500">
+                These are the cameras set up for this robot. To add or change a
+                camera, configure it on the Calibration page.
+              </p>
             </div>
 
             <Collapsible className="space-y-4 group">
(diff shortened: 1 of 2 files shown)


nicolas-rabault (approved):
Nice PR as always.
I had separate camera configuration to make a difference between cameras attached to the robot and cameras dependent of the setup the robot is used into. But I agree with you it's more clean to only have it at one location.
Thank you

PR #1, opened 2026-05-12
Patch vulnerable frontend dependencies and lockfile

arpitjain099:
## Summary
- upgrade vulnerable direct frontend deps (`vite`, `react-router-dom`, `postcss`, `lovable-tagger`)
- add targeted frontend `overrides` for vulnerable transitive packages and refresh `frontend/package-lock.json`
- apply `npm audit fix --package-lock-only` to pull in patched transitive versions

## Validation
- `cd frontend && npm audit` (0 vulnerabilities)
- `cd frontend && npm run build`

PR #68, opened 2026-07-15
feat(datasets): browse a local dataset's episodes in-app

JeremyHX:
## The gap

A dataset that exists on disk can't currently be viewed in LeLab.

The viewer is a `window.open` deep-link to the hosted [`lerobot/visualize_dataset`](https://huggingface.co/spaces/lerobot/visualize_dataset) Space, which renders from a **Hub repo path**. `Landing.tsx` encodes the consequence: `handlePickExisting` routes `source: "local" | "both"` datasets to `/upload` instead. So the dataset you just recorded has to be pushed to the Hub before you can look at it — which is backwards, since checking it is usually *why* you'd decide whether to keep it.

`/edit-dataset` has been a routed 15-line placeholder ("This page is under construction") with no backend. This fills it in.

## What this adds

Pick a dataset, page through its episodes, and watch any camera frame by frame — with a film strip and an aggregate joint-motion trace, entirely against local files.

| | |
|---|---|
| `GET /dataset-episodes` | episode list: length, duration, tasks |
| `GET /dataset-episode` | one episode: per-camera video windows, neighbours |
| `GET /dataset-frame` | one decoded frame as PNG |
| `GET /dataset-thumbnails` | film-strip thumbnails, decoded in a single pass |
| `GET /dataset-motion` | per-frame aggregate joint motion |
| `GET /dataset-video` | the camera's mp4, `inline` + Range-capable |

The Landing picker now opens local datasets in the browser instead of diverting to `/upload`. Hub-only datasets still hand off to the Space — there's nothing local to decode. **This changes existing behaviour**, so it's the piece most worth your opinion; Upload stays reachable from the browse page.

## No new dependencies

PyAV, numpy, pyarrow and Pillow all arrive via `lerobot[core_scripts,feetech,training]`, which `pyproject.toml` already requires (`core_scripts` → `dataset` → `av-dep`). Nothing was added.

`episode_media.py` reads the v3.0 layout (`meta/episodes/*.parquet` + `videos/`) directly and decodes in-process with PyAV. It deliberately avoids `LeRobotDataset`, so browsing a dataset never constructs a robot config or imports torch — listing and previewing stay cheap.

## Two details that are easy to get wrong

**Episodes share an mp4.** Locating one means resolving the file *and* its `from_/to_timestamp` window inside it. `/dataset-video` is therefore addressed by **chunk/file rather than by episode**, so the URL stays byte-identical as you page between episodes: the player seeks instead of reloading, and playback survives the switch. An episode-addressed URL hands the browser a new identifier for the same bytes, forcing a reload and silently dropping playback on every click. A test pins this.

**`duration_s` is `length / fps`, not `to_timestamp - from_timestamp`.** An episode accepted early stops its parquet at accept while the camera ran to the configured limit, so the window over-reports the demonstration (a ~2.6s episode reads as the configured 10s). The window is still exposed for the transport.

## Testing

`223 passed` (187 existing + 36 new). `ruff check`, `ruff format`, `tsc` and `eslint` are clean, and `npm run build` passes.

The new tests build a **real** (tiny) v3.0 dataset on disk — a genuine PyAV-encoded mp4 plus parquet metadata — rather than mocking the readers. The whole point of the module is that it agrees with the on-disk format, which a mock can't check. Covered: path traversal, camera/episode resolution, the packed and pre-packing layouts, the `duration_s` rule above, decode-lands-in-the-right-window, `inline` + Range on the video route, and the shared-URL contract.

Verified by hand against two datasets: [`lerobot/svla_so101_pickplace`](https://huggingface.co/datasets/lerobot/svla_so101_pickplace) (50 episodes, 30fps, cameras `side`/`up`) and an unrelated in-house one (20fps, three cameras, 7-DoF), so the format assumptions aren't tuned to one producer.

`frontend/dist` is intentionally **not** included — `build_frontend.yml` rebuilds it on merge to main.

## Documentation

`README.md` gains a **Browse** row; `CLAUDE.md` gains `datasets.py` and `episode_media.py` in the feature-module list, including the chunk/file addressing rationale. A stale comment in `server.py` ("Replay is rendered by the embedded `visualize_dataset` Space") is corrected — it's a link-out, not an embed, and local datasets no longer go there.

Unrelated and **not** touched: the README's **⏪ Replay — "Re-run any recorded episode"** row looks stale. `ba0ddfb` and `8a72410` merged record/replay and dropped the `/replay-dataset` route, so nothing re-runs an episode on hardware any more. Happy to fix in a separate PR if you agree.

## Open question

`lelab --dev` hardcodes `:8000`/`:8080` with no override, which collides with anything already on `:8000`. I worked around it locally with a `--port` uvicorn and the `?api=` override. Worth a flag? Separate PR if so.

---

Contributed by [VibeCuisine](https://github.com/VibeCuisine). Draft — happy to adjust the Landing behaviour change, the route naming, or the scope before this goes green.


nicolas-rabault:
Nice work!
This is thorough and the tests-against-real-mp4 approach is the right call. One thing to fix before merge, a copyright tweak, plus a note on your two follow-up offers.

**Thumbnail index/png misalignment**

In `extract_thumbnails`, a frame that decodes to `None` is skipped with continue, so `pngs` can come back shorter than the requested `indices`. But handle_episode_thumbnails pairs them positionally with `zip(indices, pngs, strict=False)`. If a frame in the middle of the strip fails to decode, every thumbnail after it shifts up one slot, so a tile labeled "frame 8" shows a later frame's image, clicking it seeks the player somewhere the picture never showed, and the strip silently ends short of the last frame. No error, `success:true`, valid PNGs, purely a wrong-image-under-label bug, and it surfaces exactly on a corrupt/truncated recording, which is when someone's inspecting the strip most carefully.

Suggested fix: keep the index bound to its frame instead of relying on positional alignment. e.g. have `extract_thumbnails` return `(index, png)` pairs (or `None` placeholders for misses) and build the response from those, so a dropped frame can't shift the rest.

**Copyright header**

`episode_media.py` carries `Copyright 2026 VibeCuisine` while the rest of the repo is `The HuggingFace Inc. team`. Let's merge the two rather than pick one:


`# Copyright 2026 VibeCuisine and The HuggingFace Inc. team. All rights reserved.`

**Your two follow-up PR offers — both agreed**

 - README ⏪ Replay row being stale (record/replay merged, /replay-dataset gone): yes, it's stale, go ahead and drop/fix it in its own PR.
 - lelab --dev hardcoding :8000/:8080: a --port override is worth having. A separate PR is the right scope.

Everything else, path-traversal guard, chunk/file video addressing, the duration_s = length/fps rule, sync def routes keeping PyAV decode off the event loop, looks very good.

JeremyHX:
Sounds good!
Thumbnail bug fixed in 0c9c993 with unit + route regression tests, copyright merged, and both follow-ups are coming as their own PRs

lelab/datasets.py
```diff
@@ -160,3 +161,186 @@ def list_all_datasets() -> list[dict[str, Any]]:
...
+    }
+
+
+def load_episode_frame_png(
+    repo_id: str, episode_index: int, camera: str | None, frame_index: int, max_width: int | None
+) -> bytes:
+    """One decoded frame as PNG bytes. Raises on any miss; the route maps to 404."""
+    dataset_dir = episode_media.resolve_dataset_dir(repo_id)
+    location = episode_media.locate_episode_video(dataset_dir, episode_index, camera=camera)
+    return episode_media.extract_frame_png(location, frame_index, max_width=max_width)
```
nicolas-rabault, line 282: `extract_frame_png` computes `from_timestamp + frame_index/fps`, and `_decode_at` returns the last decoded frame on overshoot. On a packed mp4 a frame index past the episode end returns the *next* episode's footage with a 200.

Repro against your own fixture (`episodes=[(0, 12), (1, 8), (2, 15)]`):

```
ep0 window=0.0..1.2
ep0 frame11 pixel=33   ep0 frame14 pixel=40   ep1 frame2 pixel=40
```

Frame 14 of ep0 is byte-identical to frame 2 of ep1. `frame_index=10_000_000` also returns 200.

The viewer clamps to `detail.length - 1` so the UI is safe, but the route is public. Suggest:

```python
rows = episode_media.read_episode_index(dataset_dir)
row = next((r for r in rows if r.episode_idx == episode_index), None)
if row and row.length and not 0 <= frame_index < row.length:
    raise episode_media.EpisodeNotFoundError(f"Frame {frame_index} out of range")
```

Wants a test. `test_frame_route_404s_on_unknown_episode` only covers a bad episode index.

frontend/src/lib/datasetApi.ts
```diff
@@ -0,0 +1,190 @@
...
+    repo_id: repoId,
+    camera,
+    chunk: String(chunk),
+    file: String(fileIndex),
+  });
+  return `${baseUrl}/dataset-video?${qs}`;
+}
+
+/** URL for a single decoded frame as a PNG. Goes into <img src>. */
+export function frameUrl(
```
nicolas-rabault, line 166: No call sites. The viewer uses `videoUrl` and `getThumbnails`, and `/dataset-frame` is only hit by tests.

Either give it a use (poster frame while the video is paused is the obvious one) or drop the helper, the route and their tests. If it stays, the out-of-range issue needs fixing first.

nicolas-rabault (changes requested):
Solid work, 

I just Requesting changes on two coupled points, both on /dataset-frame.

lelab/datasets.py
JeremyHX: Reproduced, addressed, added test. There are two checks now: the route rejects anything past the row's length, and `extract_frame_png` itself refuses targets outside the episode's video window, which also covers rows with no usable `length` where the route check never fires. Negative indices always 404. Your frame-14 repro is pinned in `test_frame_route_404s_past_episode_end`. While throwing bad inputs at the route I also hit two 500s (negative `max_width`, NUL byte in `repo_id`), fixed in the same commit.

frontend/src/lib/datasetApi.ts
JeremyHX: Sounds good, included the poster frame in this PR. Both viewer `<video>`s now use the episode's first frame as their poster, so the box shows the scene while the mp4 downloads instead of sitting black. One decode per episode+camera, browser-cached.

lelab/datasets.py
```diff
@@ -160,3 +161,198 @@ def list_all_datasets() -> list[dict[str, Any]]:
...
+# Hub-only ones to the visualize_dataset Space.
+
+
+def handle_list_episodes(repo_id: str) -> dict[str, Any]:
+    """Episode listing for one local dataset: per-episode length, duration, tasks."""
+    try:
+        dataset_dir = episode_media.resolve_dataset_dir(repo_id)
+        info = episode_media.read_info(dataset_dir)
+        cameras = episode_media.list_cameras(dataset_dir)
+        rows = episode_media.read_episode_index(dataset_dir)
```
nicolas-rabault, line 180: `read_episode_index` returns `[]` when `meta/episodes/` is missing, and this then reports success:

```json
{"success": true, "total_episodes": 4, "total_frames": 120, "cameras": ["top"], "episodes": []}
```

`total_episodes` comes from `info.json`, so the response claims 4 episodes and ships an empty list.

That state is not hypothetical. It is what a recording interrupted before `LeRobotDataset.finalize()` leaves on disk, `info.json` counts the episodes but the index is never written, and it is the reason `lelab/dataset_repair.py` exists on main. I have datasets in my own cache that were in exactly this shape. It is also the case this feature is most for: you kill a run, and the first thing you want is to look at what survived.

Blocking part is only the lie. Do not return `success: true` for an index that could not be read while `info.json` says there are episodes, say what happened so the UI can tell the user:

```python
rows = episode_media.read_episode_index(dataset_dir)
if not rows and info.get("total_episodes"):
    return {
        "success": False,
        "message": f"{repo_id} has no episode index: the recording was interrupted before it finished writing.",
    }
```

Calling `dataset_repair.repair_local_dataset(repo_id)` here to rebuild the index would be better still, but it imports lerobot's `load_episodes` and would cost you the torch-free listing path you designed for, so I am not asking for it in this PR. Your call. A test with `meta/episodes/` removed is what pins whichever you pick.

nicolas-rabault (changes requested):
Re-read the whole thing at 51801ca. Both points from last round are properly closed.

/dataset-frame is bounded twice now, and the second check (the video window, for rows with no usable length) is the one I would have asked for next, so good that you got there first. Your frame-14 repro 404s, frame 11 still decodes. frameUrl earning its keep as the poster is a better outcome than deleting the route. The two 500s you turned up while fuzzing and the center-frame seek fix were not asked for and both improve this.

On the open question in your description: the routing change is the behaviour I want. Clicking a local dataset should open the episode browser, and upload belongs as an action inside that view rather than as its own page. Settled, nothing for you to do there.

One blocker before I merge, in the review. A recording interrupted before finalize() leaves info.json counting its episodes with no meta/episodes/ written, and browse answers success: true with an episode count and an empty list. That is the case this feature is most for, so it should not report success.

Separately, heads up that PR #95 registers its own /dataset-episodes route. Local episode browsing is yours, this PR is where it lives, so no action needed on your side.

ravediamond:
Hi @JeremyHX — any update on this? Want a hand finishing it up?

ravediamond:
Hello @nicolas-rabault,

JeremyHX hasn't responded in a while, so I pushed a fix for the remaining blocker (missing episode index reporting `success: true`) to my fork: https://github.com/VibeCuisine/leLab/pull/1.

Since I don't have write access to push directly onto his branch, could you pull it in — either by merging that PR or cherry-picking the commit — so this can move forward?

lelab/datasets.py
```diff
@@ -160,3 +161,198 @@ def list_all_datasets() -> list[dict[str, Any]]:
...
+    except episode_media.DatasetNotFoundError as e:
+        return {"success": False, "message": str(e)}
+    except Exception as e:
+        logger.warning(f"Could not list episodes for {repo_id}: {e}")
+        return {"success": False, "message": f"Could not read dataset: {e}"}
+
+    return {
```
nicolas-rabault: Still the spot. 16e4644 on VibeCuisine/leLab#1 inserts the guard right here:

```python
if not rows and info.get("total_episodes"):
    return {
        "success": False,
        "message": f"{repo_id} has no episode index: the recording was interrupted before it finished writing.",
    }
```

I ran it both ways: `test_list_episodes_reports_failure_when_index_missing` fails at this head and passes with the commit, and the full suite is green with it. Landing that commit on this branch closes this thread.

PR #72, opened 2026-07-17
fix: Windows stability issues (console encoding, camera enumeration, …

### LINKED ISSUE #78: Recording session randomly stops in LeLab with no error shown in GUI
When starting a training/recording session, LeLab sometimes stops recording
partway through with no error message in the GUI. The backend log shows the
camera and arm disconnecting and the session being marked as failed, but no
underlying exception/traceback is printed or surfaced to the frontend. This is
happening on the most recent version of LeLab, on Windows 11.

Steps to reproduce

Start a recording session in LeLab. Specify episode number (i.e: 50).
Recording runs fine, but intermittently stops early (i.e: episode 13.
GUI shows no error. Status just goes to "failed" silently.
Logs:

INFO 2026-07-17 11:49:33 a_opencv.py:601 OpenCVCamera(0) disconnected.
INFO 2026-07-17 11:49:33 a_opencv.py:601 OpenCVCamera(1) disconnected.
INFO 2026-07-17 11:49:33 follower.py:240 so-arm101 SOFollower disconnected.
INFO 2026-07-17 11:49:33 o_leader.py:163 so-arm101 SOLeader disconnected.
ERROR 2026-07-17 11:49:33 b\record.py:312 Recording session failed
Svt[info]: Only a single picture was passed in. Consider setting avif=1 for improved efficiency and reduced memory usage.
Svt[info]: Only a single picture was passed in. Consider setting avif=1 for improved efficiency and reduced memory usage.
INFO 2026-07-17 11:49:34 b\record.py:327 Recording session ended
INFO 2026-07-17 11:49:34 b\record.py:404 📡 RECORDING STATUS REQUEST: Session failed with error - frontend should stop polling
📡 STATUS CHANGE: Frontend is still polling after session error - should stop now

ZouzouWP:
## Related issue

Fixes #78 — the reported symptoms match exactly: camera/arm disconnecting mid-session, `Svt[info]: Only a single picture was passed in` in the logs, and the session silently marked "failed" with nothing surfaced to the GUI. Fixes 2 (camera enumeration) and 3 (SVT-AV1 encoder starving the motor-bus read loop) address the root cause; fix 4 (surfacing the real exception) closes the "no error shown in GUI" part.

Four separate crashes that only show up on Windows with real hardware attached. Grouped in one PR because they were all found the same week, not because they're related — happy to split if preferred.

## 1. Console encoding crash (cp1252)

**Symptom:** an emoji in a log/print line could raise an unhandled `UnicodeEncodeError` on a Windows console using the legacy cp1252 code page, turning an endpoint into a blank 500 with no useful message reaching the frontend.

**Fix:** `stdout`/`stderr` are reconfigured with `errors="replace"` at import time, so unencodable characters degrade instead of crashing the process.

## 2. Camera list sometimes empty

**Symptom:** `/available-cameras` would intermittently return an empty or wrong list, depending on which FastAPI worker thread handled the request.

**Root cause:** `pygrabber` enumerates cameras via COM, which must be initialized per-thread. FastAPI's thread pool never did that, so enumeration silently failed on threads that hadn't.

**Fix:** explicitly initialize/uninitialize COM around the enumeration call.

## 3. Recording crashes with "no status packet"

**Symptom:** recording sessions aborted mid-way with a serial `ConnectionError`, non-deterministically, on otherwise-healthy hardware.

**Root cause:** the default video encoder (libsvtav1) is CPU-heavy enough that two parallel encode streams could starve the motor-bus read loop of CPU time.

**Fix:** default to h264/ultrafast, measured ~11x faster to encode on this hardware.

## 4. Real errors hidden behind "check logs"

**Symptom:** any recording failure showed the same generic message to the user — the real exception (exactly what let us diagnose #2 and #3 above) never left the server logs.

**Fix:** `/recording-status` now includes the actual exception message.

## Testing

Each fix corresponds to a crash reproduced and captured in server logs beforehand, then confirmed resolved after the change.

### DIFF
--- a/lelab/record.py
+++ b/lelab/record.py
@@ -23,6 +23,7 @@ from typing import Any
 from pydantic import BaseModel
 
 from lerobot.configs.dataset import DatasetRecordConfig
+from lerobot.configs.video import RGBEncoderConfig
 from lerobot.datasets import LeRobotDataset
 from lerobot.robots.so_follower import SO101FollowerConfig
 
@@ -183,6 +184,10 @@ def create_record_config(request: RecordingRequest) -> RecordConfig:
         tags=with_lelab_tag(request.tags) if request.push_to_hub else None,
         private=request.private,
         streaming_encoding=request.streaming_encoding,
+        # libsvtav1 (lerobot's default) is CPU-heavy enough on this machine to starve
+        # the motor bus read loop mid-recording ("no status packet" ConnectionError,
+        # see 2026-07-15 crashes). h264/ultrafast measured ~11x faster to encode here.
+        rgb_encoder=RGBEncoderConfig(vcodec="h264", preset="ultrafast"),
     )
 
     # Create the main record config
@@ -309,11 +314,15 @@ def handle_start_recording(request: RecordingRequest) -> dict[str, Any]:
                     "robot_type": getattr(dataset.meta, "robot_type", "Unknown robot"),
                 }
             except Exception as e:
-                logger.exception("Recording session failed")
+                logger.exception("Recording session failed: %r", e)
                 current_phase = "error"
                 if recording_start_time:
                     session_end_elapsed_seconds = int(time.time() - recording_start_time)
-                last_recording_info = {"success": False, "error": str(e)}
+                last_recording_info = {
+                    "success": False,
+                    "error": str(e) or repr(e),
+                    "dataset_repo_id": request.dataset_repo_id,
+                }
             finally:
                 if current_phase != "error":
                     current_phase = "completed"
@@ -419,7 +428,11 @@ def handle_recording_status() -> dict[str, Any]:
             "rerecord_episode": recording_active
             and current_phase == "recording",  # Only during recording phase
         },
-        "message": "Recording session failed with error - check logs"
+        "message": (
+            f"Recording failed: {last_recording_info['error']}"
+            if last_recording_info and last_recording_info.get("error")
+            else "Recording session failed with error - check logs"
+        )
         if current_phase == "error"
         else (
             "Recording session has ended - stop polling"
--- a/lelab/server.py
+++ b/lelab/server.py
@@ -36,6 +36,15 @@ from starlette.exceptions import HTTPException as StarletteHTTPException
 from starlette.responses import Response
 from starlette.types import Scope
 
+# Windows consoles often use a legacy code page (cp1252) that can't encode the
+# emoji used in log/print output; an unhandled UnicodeEncodeError inside an
+# endpoint then turns into a 500 and hides the real error from the frontend.
+# Replace unencodable characters instead of crashing.
+for _stream in (sys.stdout, sys.stderr):
+    if _stream is not None and hasattr(_stream, "reconfigure"):
+        with contextlib.suppress(Exception):
+            _stream.reconfigure(errors="replace")
+
 from . import datasets as dataset_browser
 
 # Import our custom calibration functionality
@@ -998,6 +1007,18 @@ def _windows_cameras() -> list[dict[str, Any]]:
     frontend match each index to the browser's ``MediaDeviceInfo.label`` for the
     live preview. Falls back to generic names if pygrabber is unavailable.
     """
+    # pygrabber uses COM, which must be initialized per-thread. FastAPI serves
+    # sync endpoints from a thread pool whose threads never call CoInitialize,
+    # so enumeration fails intermittently depending on which thread handles the
+    # request. Initialize COM here and balance with CoUninitialize.
+    import ctypes
+
+    com_initialized = False
+    try:
+        hr = ctypes.windll.ole32.CoInitialize(None)
+        com_initialized = hr in (0, 1)  # S_OK or S_FALSE (already initialized)
+    except Exception:
+        pass
     try:
         from pygrabber.dshow_graph import FilterGraph
 
@@ -1007,6 +1028,9 @@ def _windows_cameras() -> list[dict[str, Any]]:
         import cv2
 
         return _generic_cv2_cameras(cv2.CAP_DSHOW)
+    finally:
+        if com_initialized:
+            ctypes.windll.ole32.CoUninitialize()
     return [{"index": i, "name": name, "available": True} for i, name in enumerate(names)]
 
 


lelab/server.py
```diff
@@ -36,6 +36,15 @@
 from starlette.responses import Response
 from starlette.types import Scope
 
+# Windows consoles often use a legacy code page (cp1252) that can't encode the
+# emoji used in log/print output; an unhandled UnicodeEncodeError inside an
+# endpoint then turns into a 500 and hides the real error from the frontend.
+# Replace unencodable characters instead of crashing.
+for _stream in (sys.stdout, sys.stderr):
```
nicolas-rabault, line 43: Blocking: this statement sits between the stdlib imports and the local ones, so every `from .` import below it trips E402. `ruff check` reports 12 errors on this branch (base is clean), and `quality.yml` runs pre-commit on every PR, so CI will fail.

Suggest moving it into `main()` in `lelab/scripts/lelab.py`. Process wide stdio setup belongs at the entrypoint rather than as an import side effect, and it sidesteps E402 entirely.

lelab/server.py
```diff
@@ -36,6 +36,15 @@
...
+    if _stream is not None and hasattr(_stream, "reconfigure"):
+        with contextlib.suppress(Exception):
+            _stream.reconfigure(errors="replace")
```
nicolas-rabault, line 46: Minor: `errors="replace"` turns the emoji into `?`. Adding `encoding="utf-8"` makes them render correctly on Windows Terminal and modern conhost.

Alternatively, dropping the emoji from the log lines in `record.py` (four `📡` lines around 414-419) removes the failure mode at the source.

lelab/server.py
```diff
@@ -998,6 +1007,18 @@ def _windows_cameras() -> list[dict[str, Any]]:
...
     live preview. Falls back to generic names if pygrabber is unavailable.
     """
+    # pygrabber uses COM, which must be initialized per-thread. FastAPI serves
+    # sync endpoints from a thread pool whose threads never call CoInitialize,
+    # so enumeration fails intermittently depending on which thread handles the
+    # request. Initialize COM here and balance with CoUninitialize.
+    import ctypes
+
+    com_initialized = False
+    try:
```
nicolas-rabault, line 1017: The fix is right, including the `hr in (0, 1)` guard that correctly skips `CoUninitialize` on `RPC_E_CHANGED_MODE`.

Two things though: `except Exception: pass` swallows silently, and on macOS/Linux `ctypes.windll` raises `AttributeError` on every call (the existing tests do call `_windows_cameras()` off Windows). Suggest guarding on `sys.platform == "win32"` and logging at debug instead of swallowing.

lelab/record.py
```diff
@@ -183,6 +184,10 @@ def create_record_config(request: RecordingRequest) -> RecordConfig:
         tags=with_lelab_tag(request.tags) if request.push_to_hub else None,
         private=request.private,
         streaming_encoding=request.streaming_encoding,
+        # libsvtav1 (lerobot's default) is CPU-heavy enough on this machine to starve
+        # the motor bus read loop mid-recording ("no status packet" ConnectionError,
+        # see 2026-07-15 crashes). h264/ultrafast measured ~11x faster to encode here.
+        rgb_encoder=RGBEncoderConfig(vcodec="h264", preset="ultrafast"),
```
nicolas-rabault, line 190: Blocking, and the main discussion point. This changes the default encoder for every platform, not just Windows.

lerobot v0.6.0 already has better levers for this exact problem:

- `vcodec="auto"` probes `h264_videotoolbox` / `h264_nvenc` / `h264_vaapi` / `h264_qsv` and falls back to `libsvtav1`. Hardware encoding is near zero CPU, which is the actual root cause, and it does not degrade Mac and Linux.
- `DatasetRecordConfig.encoder_threads` caps encoder threads directly, which is the surgical fix for "encoder starves the motor bus read loop".

Risk worth checking before merge: `RGBEncoderConfig.__post_init__` calls `resolve_vcodec()`, which raises `ValueError: Unsupported video codec: h264` when the installed PyAV has no libx264 encoder. That throws inside `create_record_config`, so recording would die before starting on machines that work today. Please confirm on Linux and macOS with the pinned lerobot v0.6.0, not only on the Windows box.

Also, `crf=30` and `g=2` are inherited while `preset="ultrafast"` disables most compression tools, so datasets get noticeably larger. The PR measures encode speed (~11x) but not file size.

Last, the comment references "this machine" and "2026-07-15 crashes", which a future reader cannot verify. If the setting stays, expose it as a `RecordingRequest` field (same as `video` and `streaming_encoding` already are) or gate it on `sys.platform == "win32"`.

lelab/record.py
```diff
@@ -309,11 +314,15 @@ def recording_worker():
                     "robot_type": getattr(dataset.meta, "robot_type", "Unknown robot"),
                 }
             except Exception as e:
-                logger.exception("Recording session failed")
+                logger.exception("Recording session failed: %r", e)
```
nicolas-rabault, line 317: Nit: `logger.exception` already emits the traceback, whose last line is the exception repr, so `%r` is redundant.

lelab/record.py
```diff
@@ -309,11 +314,15 @@ def recording_worker():
...
                 current_phase = "error"
                 if recording_start_time:
                     session_end_elapsed_seconds = int(time.time() - recording_start_time)
-                last_recording_info = {"success": False, "error": str(e)}
+                last_recording_info = {
+                    "success": False,
+                    "error": str(e) or repr(e),
```
nicolas-rabault, line 323: `str(e) or repr(e)` is a good catch for exceptions with an empty message. Keep this.

lelab/record.py
```diff
@@ -309,11 +314,15 @@ def recording_worker():
...
+                    "dataset_repo_id": request.dataset_repo_id,
```
nicolas-rabault, line 324: Blocking: this is an unintended regression. `handle_get_dataset_info` (line 479) short-circuits on `last_recording_info.get("dataset_repo_id") == request.dataset_repo_id`. Before this PR failure dicts had no `dataset_repo_id`, so it never matched and the on-disk dataset was loaded. Now a session that crashed after saving N episodes returns `{"success": False, ...}` and the partial dataset is never read, which is exactly the scenario this PR targets.

Also, this dict carries `error` but no `message`, unlike the other failure return at line 498, so the frontend reads `undefined`.

Suggest gating on success:

```python
if last_recording_info and last_recording_info.get("success") \
        and last_recording_info.get("dataset_repo_id") == request.dataset_repo_id:
```

lelab/record.py
```diff
@@ -419,7 +428,11 @@ def handle_recording_status() -> dict[str, Any]:
             "rerecord_episode": recording_active
             and current_phase == "recording",  # Only during recording phase
         },
-        "message": "Recording session failed with error - check logs"
+        "message": (
```
nicolas-rabault, line 431: The global is read twice, guard then subscript, and `handle_start_recording` sets it to `None` at line 242. Narrow window, but the result is a 500 on `/recording-status`, which is the endpoint whose job is reporting errors. Bind once:

```python
info = last_recording_info
error = info.get("error") if info else None
```

Then build the message above the `status = {...}` literal. The triple nested conditional inside a dict literal is hard to follow as written.

nicolas-rabault (changes requested):
Thanks a lot for this, genuinely useful work. Chasing four separate crashes on real hardware and coming back with a root cause for each one, COM per-thread init and encoder CPU starvation in particular, is exactly the kind of debugging that is hard to do remotely. The write up made the review easy to follow.

Three things before merge:

 1. `ruff check` fails with 12 E402 errors from the stdout reconfigure placement, so Quality CI will go red. Moving the block into `lelab/scripts/lelab.py` fixes it and is a better home anyway.
 2. Adding `dataset_repo_id` to the error dict silently changes `handle_get_dataset_info`, so a partially recorded dataset can no longer be read from disk. Gating on success should sort it.
 3. The encoder change is the one I would like to discuss rather than merge as is. lerobot already exposes `vcodec="auto"` (hardware encoders with an automatic fallback) and `encoder_threads`, both of which target the CPU starvation more directly without changing behavior for Mac and Linux users. There is also a hard failure risk if the installed PyAV lacks libx264.
On splitting: yes please. The three Windows scoped fixes are low risk and can land as soon as the E402 and the dataset info regression are addressed. The encoder change deserves its own PR with file size numbers next to the encode speed numbers, since it alters what gets stored and pushed to the Hub.

A few tests would help too, and the suite already covers these exact functions. A test that `handle_get_dataset_info` still loads a partial dataset after a failed session would have caught point 2.

Really appreciate the effort here, happy to help with the encoder follow up if useful.

lelab/server.py
nicolas-rabault, line 43: Still blocking. Verified with the pinned ruff 0.14.1 from `.pre-commit-config.yaml`: `ruff check lelab/` gives 12 E402 here, `All checks passed!` on `main`, so `quality.yml` goes red.

Fix is still to move it into `main()` in `lelab/scripts/lelab.py`. `tests/test_scripts_lelab.py` already drives that entrypoint.

This hunk also conflicts now: the import below is `from . import datasets as dataset_browser, record as _record` on `main`.

lelab/server.py
```diff
@@ -998,6 +1007,18 @@ def _windows_cameras() -> list[dict[str, Any]]:
     frontend match each index to the browser's ``MediaDeviceInfo.label`` for the
     live preview. Falls back to generic names if pygrabber is unavailable.
     """
+    # pygrabber uses COM, which must be initialized per-thread. FastAPI serves
```
nicolas-rabault, line 1010: Superseded, drop this from the split.

#99 patches the same bug in this same function through a `_windows_com_initialized()` context manager over comtypes' `CoInitialize` / `CoUninitialize`, with tests in `tests/test_server.py`. Only one can land and #99 is the one: smaller, library API instead of `ole32` directly, logs on failure rather than `except Exception: pass`, and its `ImportError` path no-ops off Windows where `ctypes.windll` raises `AttributeError` in the existing tests.

You found the bug first and the root cause was right, `hr in (0, 1)` included. It is the code that is duplicated, not the diagnosis.

lelab/record.py
nicolas-rabault, line 190: Own PR, per the split. Three things checked against the pinned lerobot v0.6.0 since July:

- `vcodec="auto"` works as advertised: on macOS it resolves to `h264_videotoolbox`. `encoder_threads` is at `lerobot/configs/dataset.py:73`.
- The hard-failure risk I raised did not reproduce. `RGBEncoderConfig(vcodec="h264", preset="ultrafast")` constructs fine on macOS with the pinned build. `resolve_vcodec()` does still raise `ValueError: Unsupported video codec` when PyAV lacks libx264, so please confirm on Linux, but treat my July note as unproven rather than as a known break.
- `crf=30` and `g=2` are confirmed inherited next to `preset="ultrafast"`, so the file size question stands.

The comment needs to lose "this machine" and "2026-07-15 crashes" whatever setting lands: a future reader cannot check either.

lelab/record.py
nicolas-rabault, line 317: Drop on the rebase. `main` deliberately went the other way, because lerobot's `init_logging()` installs a root formatter that renders the message only:

```python
logger.error("Recording session failed: %s\n%s", e, traceback.format_exc())
```

`logger.exception` here loses the traceback again.

lelab/record.py
nicolas-rabault, line 323: Keep this one. `main` has `"error": str(e)`, so `or repr(e)` is the whole remaining delta and it earns its place for exceptions that stringify to nothing.

lelab/record.py
nicolas-rabault, line 324: My July blocking point is resolved, and not by the gating I suggested, so drop this line on the rebase: `main` already writes both `dataset_repo_id` and `saved_episodes` here on purpose, and the short-circuit in `handle_get_dataset_info` is how `frontend/src/pages/Upload.tsx` learns how many episodes a crashed session saved.

lelab/record.py
nicolas-rabault, line 431: This half has not landed, `main` still returns the generic "check logs", so it is the piece I would take first.

Two carry-overs. Bind the global once, since `handle_start_recording` sets `last_recording_info` to `None` at line 242, so the guard can pass and the subscript still raise on the one endpoint whose job is reporting errors:

```python
info = last_recording_info
error = info.get("error") if info else None
```

Then build the message above the `status = {...}` literal instead of nesting a second conditional inside it. `tests/test_record.py` already has `test_recording_status_handler_exposes_state_fields` for the assertion to sit next to.

nicolas-rabault (changes requested):
Thanks for chasing four real crashes on hardware nobody else here has.

The ask: rebase, then split as you offered. Drop the COM hunk, #99 covers it. The encoder wants its own PR. Details inline, including one correction to my July note that goes in your favour.

PR #81, opened 2026-07-18
feat: add OMX-AI robot arm support alongside SO-101

### LINKED ISSUE #79: Feature Request: Support for OMX-AI in leLab
## Describe the Feature Request
Provide native support for the **OMX-AI** robotic arm in the `leLab` web GUI, alongside the current SO-100 series.

Currently, the underlying `lerobot` library officially supports Dynamixel motors and the OMX-AI platform. However, `leLab` restricts connection and usage to the hardcoded SO-101 leader/follower arms. Introducing abstraction for robot configurations will allow users of OMX-AI to also benefit from leLab's zero-code web UI for calibration, teleoperation, recording, and inference.

## Current Limitations / Code References
Here are the areas in the codebase that are currently hardcoded for SO-101:

### 1. Backend Configurations (`lelab`)
*   **Teleoperation & Verification**: In `lelab/teleoperate.py` (lines 23–24), `SO101FollowerConfig`, `SO101Follower`, `SO101LeaderConfig`, and `SO101Leader` are directly imported. They are instantiated directly in lines 169–184.
*   **Dataset Recording**: In `lelab/record.py` (lines 27–31), `SO101FollowerConfig` and `SO101LeaderConfig` are imported and instantiated on lines 159–166.
*   **Calibration**: In `lelab/calibrate.py` (lines 256–263), calibration configuration setup specifically expects `SO101FollowerConfig` and `SO101LeaderConfig`.
*   **Inference Rollouts**: In `lelab/rollout.py` (line 284), `--robot.type=so101_follower` is hardcoded as an argument for the rollout subprocess.

### 2. Frontend Calibration Configuration (`frontend`)
*   **Joint Calibration Targets**: In `frontend/src/lib/calibrationTargets.ts` (lines 4–25), the target calibration limits are explicitly hardcoded for `SO101_LEADER_TARGETS` and `SO101_FOLLOWER_TARGETS` (e.g., `shoulder_pan`, `shoulder_lift`, `elbow_flex`, `wrist_flex`, `wrist_roll`, `gripper`).

---

## Proposed Solutions
1. **Dynamic Robot Configuration Loading**:
   *   Refactor the backend (e.g., `teleoperate.py`, `record.py`, `calibrate.py`) to dynamically import and build configuration classes depending on the target robot specified (e.g., loading `OMXAIFollowerConfig` / `OMXAILeaderConfig` if `omx-ai` is chosen).
2. **Abstract Calibration Targets on the Frontend**:
   *   Expose the calibration joint targets dynamically from the backend (by reading properties from the respective `lerobot` robot class) or refactor the frontend configuration to support mappings based on the active robot configuration profile.
3. **Parameterize Rollouts**:
   *   Allow the frontend or user configuration file to dynamically set the `--robot.type` parameter instead of hardcoding it in `rollout.py`.

robertchoi:
Closes #79

## Summary

This PR adds native support for the ROBOTIS OMX-AI robotic arm to LeLab's web GUI, alongside the existing SO-101 leader/follower arms. All four hardcoded areas identified in #79 are now abstracted behind a per-robot `robot_type` setting.

### Backend: dynamic robot configuration loading
- New `make_device_config()` / `make_device()` helpers in `lelab/utils/devices.py` dispatch LeRobot config and device classes by `robot_type` (`so101` → `SO101Follower`/`SO101Leader`, `omx_ai` → `OmxFollower`/`OmxLeader`).
- `teleoperate.py`, `record.py`, and `calibrate.py` build devices through these helpers instead of importing SO-101 classes directly; every request schema gains a `robot_type` field (default `"so101"`, fully backward compatible).
- Calibration file paths (`utils/config.py`) resolve per robot type (`so_leader`/`so_follower` vs `omx_leader`/`omx_follower`), matching the directories LeRobot derives from each device's `name`.
- Calibration picks the `POSITION` operating-mode value per bus protocol: Feetech uses `0`, but on Dynamixel (OMX) `0` means CURRENT mode — writing the Feetech value would have put OMX motors in current-control mode.
- OMX-AI arms are teleoperable without a range calibration (matching upstream LeRobot's factory-default calibration for OMX).

### Rollout parameterization
- `rollout.py` resolves `--robot.type` from the request (`so101_follower` / `omx_follower`) instead of hardcoding it.

### Frontend
- Robot records store a `robot_type`; it is sent with every teleoperation, recording, inference, and calibration start request.
- Robot model can be chosen at creation time (SO-101 / OMX-AI chips in the robot selector) and changed later on the Configure page.
- `calibrationTargets.ts` gains OMX-AI per-joint target ranges, selected by the active robot's type.
- 3D viewer support for the OMX follower (URDF + STL meshes, Apache-2.0 licensed from ROBOTIS).

## Testing
- New regression test asserting the operating-mode value matches the bus protocol (Feetech `0` vs Dynamixel `3`).
- Teleoperation tests now patch the `make_device` seam, so both SO-101 and OMX paths are testable.
- Full suite: 188 passed; `ruff check` clean.
- Manually verified on Windows: robot creation with OMX-AI type persists `robot_type: "omx_ai"`, and SO-101 flows are unchanged.

## Notes for reviewers
- `robot_type` defaults to `"so101"` everywhere, so existing robot records and API clients keep working unchanged.
- Adding further robot models now only requires a new branch in `make_device_config`/`make_device`, a calibration-path folder mapping, and (optionally) frontend target ranges.


lelab/utils/devices.py
```diff
@@ -49,3 +49,95 @@ def _force_close_device_resources(device: Any, logger: logging.Logger) -> None:
...
+    robot_type: str,
+    side: Literal["leader", "follower"],
+    port: str,
+    config_id: str,
+    cameras: dict | None = None,
+) -> Any:
+    """Create a LeRobot device config object dynamically based on robot type."""
+    model = robot_type.lower()
+
+    if "so" in model:
```
nicolas-rabault, line 93: This is the shape the split is about, so treating it as the anchor for that ask. Two concrete things a registry fixes that this cannot: `"so" in model` is substring matching and it is tested before `omx`, so any future arm whose name contains those two letters silently resolves to SO-101; and `make_device_config` and `make_device` carry duplicate branch trees that have to be kept in sync by hand forever.

One entry keyed on an exact `robot_type` string, holding the config class and the device class together, removes both and gives you a real lookup failure on an unknown type instead of a silent fallback.

lelab/calibrate.py
```diff
@@ -27,22 +27,28 @@
...
 logger = logging.getLogger(__name__)
 
 
+def _position_operating_mode(bus) -> int:
+    """POSITION operating-mode register value for the bus's motor protocol.
+
+    Feetech and Dynamixel disagree on the value (0 vs 3); writing the Feetech
+    value to a Dynamixel bus (OMX) would put the motors in CURRENT mode.
+    """
+    if "dynamixel" in type(bus).__module__:
```
nicolas-rabault, line 45: Blocker, and it breaks the arm this PR is adding. Picking the right `OperatingMode` value here is necessary but not sufficient: `_complete_calibration` further down still writes `drive_mode=0` for every motor. I checked `OmxLeader.calibrate()` at the pinned `v0.6.0` and it sets `drive_mode=1` and `homing_offset=100` on the gripper.

So running the newly exposed OMX wizard saves a leader calibration that inverts the gripper against what lerobot expects. Either keep OMX off the wizard entirely, or make the calibration recipe part of the arm entry rather than a constant in `_complete_calibration`.

lelab/utils/config.py
```diff
@@ -31,6 +31,27 @@
...
+    """Get the directory where calibration configurations are stored for a given robot model."""
+    base_config_path = LEADER_CONFIG_PATH if side == "leader" else FOLLOWER_CONFIG_PATH
+    base = os.path.dirname(base_config_path)
+    model = (robot_type or "so101").lower().replace("-", "_")
+    if "so" in model:
+        folder = f"so_{side}"
+    elif "omx" in model:
+        folder = f"omx_{side}"
+    else:
+        folder = f"so_{side}"
```
nicolas-rabault, line 52: Blocker: this `else` silently resolves an unknown `robot_type` to the SO-101 directory, and `DELETE /calibration-configs/{device_type}/{config_name}` reaches this function. A typo in `robot_type` on that request deletes a real SO-101 calibration while the caller believes it asked about another arm. Please raise on an unknown type instead.

Separately, the folder names themselves are right: I checked against lerobot v0.6.0 and `OmxLeader.name` is `omx_leader`, `OmxFollower.name` is `omx_follower`, so lerobot will find the files where you put them.

lelab/rollout.py
```diff
@@ -271,17 +245,26 @@ def handle_start_inference(request: InferenceRequest) -> dict[str, Any]:
...
-        follower_id = setup_follower_calibration_file(request.follower_config)
+        follower_id = setup_follower_calibration_file(request.follower_config, request.robot_type)
         policy_path = _resolve_policy_path(request.policy_ref)
 
+        # Resolve robot type argument for lerobot CLI
+        model = request.robot_type.lower()
+        if "so" in model:
+            robot_type_arg = "--robot.type=so101_follower"
+        elif "omx" in model:
+            robot_type_arg = "--robot.type=omx_follower"
```
nicolas-rabault, line 256: Blocker, same class as the one in `get_calibration_path` but with worse consequences: the `else` sends a mistyped `robot_type` to `so101_follower`, which means a policy runs on the wrong physical arm. Raise instead.

The `omx_follower` string itself is correct, I verified `OmxFollowerConfig` carries `@RobotConfig.register_subclass("omx_follower")` at v0.6.0.

frontend/public/omx-urdf/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
nicolas-rabault, line 1: Blocker on the redistribution, though the answer is an easy one. The licensing is fine: I traced these to `ROBOTIS-GIT/open_manipulator` (`open_manipulator_description/`, `omx_f.urdf.xacro`), that repo is Apache-2.0, and leLab is Apache-2.0 too, so they are compatible.

What is missing is attribution. This file is the unmodified Apache boilerplate with `Copyright [yyyy] [name of copyright owner]` still a literal placeholder, and nothing in the branch records where the meshes came from. Apache-2.0 section 4 asks for the original copyright notice to be retained, so we cannot ship them as-is. Please fill in the ROBOTIS copyright line and add the source repo, path, and revision alongside it.

nicolas-rabault (changes requested):
Thanks Robert, this is a very good write-up and more arms in LeLab is something we need.

One ask: split it in two, a robot registry with SO-101 as the only entry and no behaviour change, then OMX-AI as one entry on top. The call sites are listed in the review, so PR 1 is already scoped.

Five blocking notes inline, all in the OMX half. Two of the bot's comments are wrong, I have said which.

I have SO-101 but no OMX, so I will cover the regression side and will need you to show the OMX half working. What I need is in the review.

PR #87, opened 2026-07-25
fix(update): only flag an update when the install is behind

cn0303:
## What does this PR do?

The update notifier flagged "update available" whenever the repo HEAD differed
from the installed commit, in **either** direction. So an install that was ahead
of the default branch (a dev build) or on a branch that had diverged (a
force-push or rewritten history) still got nagged to update, and the in-place
`--force` reinstall the button offers would have discarded those local commits.

## The cause

`_compute_status()` compared the installed commit to the repo HEAD and set
`update_available = True` for any difference:

```python
compare = _github_json(f"/repos/{owner}/{repo}/compare/{current}...{latest}")
if isinstance(compare, dict):
    base.commits_behind = compare.get("ahead_by")
    base.compare_url = compare.get("html_url")
base.update_available = True   # <- true for ahead / behind / diverged alike
```

GitHub's `compare(base=installed, head=latest)` already classifies the
relationship in its `status` field (`ahead`, `behind`, `diverged`, `identical`),
but the code ignored it.

## The fix

Gate on `status == "ahead"`, the one case where the repo has strictly newer
commits the install can fast-forward to:

```python
if isinstance(compare, dict):
    base.commits_behind = compare.get("ahead_by")
    base.compare_url = compare.get("html_url")
    base.update_available = compare.get("status") == "ahead"
```

`behind` (the install is ahead of the default branch) and `diverged` now stay
silent, as does an unavailable compare (for instance when the installed commit is
no longer in the repo), rather than guessing.

## Before / after

| Install relationship to default branch | compare status | Before | After |
|---|---|---|---|
| Behind (normal, new commits upstream) | `ahead` | "update available" | "update available" (unchanged) |
| Up to date | `identical` | silent | silent (unchanged) |
| Ahead (dev build) | `behind` | **"update available", 0 commits behind** | silent |
| Diverged (force-push / rewritten history) | `diverged` | **"update available"** | silent |
| GitHub compare unreachable | — | **"update available"** | silent |

## Testing

Reproduced with unit tests that feed `_github_json` a canned compare response.
Three new tests (`diverged`, install-ahead, compare-unavailable) fail against the
current code and pass with the fix; the existing "update available" and
"up to date" cases still pass. `pytest tests/test_update.py`: 20 passed. ruff
check and ruff format green; pre-commit (mypy, bandit, typos) green.

### DIFF
--- a/lelab/update.py
+++ b/lelab/update.py
@@ -185,11 +185,19 @@ def _compute_status() -> dict[str, Any]:
         base.commits_behind = 0
         return base.model_dump()
 
+    # compare(base=installed, head=latest) reports how the repo relates to the
+    # installed commit. Only status "ahead" means the repo has strictly newer
+    # commits the install can fast-forward to — the one case worth a nudge.
+    # "behind" (the install sits ahead of the default branch) and "diverged"
+    # (history rewritten / force-push) must not nag: the in-place --force
+    # reinstall would throw away those local commits. If the compare is
+    # unavailable (e.g. the installed commit is no longer in the repo), stay
+    # silent rather than guess.
     compare = _github_json(f"/repos/{owner}/{repo}/compare/{current}...{latest}")
     if isinstance(compare, dict):
         base.commits_behind = compare.get("ahead_by")
         base.compare_url = compare.get("html_url")
-    base.update_available = True
+        base.update_available = compare.get("status") == "ahead"
     return base.model_dump()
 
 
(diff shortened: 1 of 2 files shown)


nicolas-rabault (approved):
Good catch, and the diagnosis is exactly right: the old code treated any
difference as "behind", which nags dev builds and branch installs for no
reason.

Merging.

PR #88, opened 2026-07-25
refactor(calibrate): name the motor-position numbers

cn0303:
## What does this PR do?

`calibrate.py` decided "is this a plausible motor-position reading" with bare
literals repeated in three places, plus two more single-use thresholds:

- `pos <= 0 or pos >= 5000` (skip invalid) in `get_status`
- `pos > 0 and pos < 5000` (keep valid) in `_step_range_recording`, twice
- `abs(pos - prev) > 2000` : the encoder wrap-around jump
- `range_diff < 100` : the "joint barely moved" warning threshold

Same concepts written slightly differently, so changing the valid range meant
finding every copy and hoping they agreed.

## The change

Hoist the values into named module constants and route the three plausible-range
checks through one helper:

```python
_MIN_VALID_POSITION = 0
_MAX_VALID_POSITION = 5000
_MAX_POSITION_JUMP = 2000
_MIN_CALIBRATION_RANGE = 100


def _is_valid_position(pos: float) -> bool:
    return _MIN_VALID_POSITION < pos < _MAX_VALID_POSITION
```

The three filter sites become `_is_valid_position(pos)` / `not
_is_valid_position(pos)`, and the two remaining thresholds read
`_MAX_POSITION_JUMP` / `_MIN_CALIBRATION_RANGE`.

## No behaviour change

This is a pure refactor. The helper's bounds are exclusive on both ends, which
matches the old `pos > 0 and pos < 5000` exactly (and its negation `pos <= 0 or
pos >= 5000`). The wrap-around and minimum-range thresholds keep their exact
values.

## Testing

Behaviour-preserving, so the proof is the existing calibrate suite staying green
plus a new parametrised boundary test on `_is_valid_position` pinning the
exclusive bounds (0 and 5000 both invalid; 1, 4095, 4999 valid). `pytest
tests/test_calibrate.py`: 15 passed. Full suite: 207 passed (the only 3 failures
are pre-existing on main, in dataset_repair/jobs, unrelated). ruff check and ruff
format green; pre-commit (mypy, bandit, typos) green.

There is no hardware before/after to run here; the values and comparisons are
unchanged, only their names.

### DIFF
--- a/lelab/calibrate.py
+++ b/lelab/calibrate.py
@@ -42,6 +42,28 @@ from .utils.devices import safe_disconnect_device
 
 logger = logging.getLogger(__name__)
 
+# Feetech sts3215 Present_Position readings are 12-bit (0-4095). A reading at or
+# below 0, or at/above this loose ceiling, is a bad frame (disconnected motor or
+# an encoder wrap-around) rather than a real joint position, so it's filtered out
+# everywhere a raw position is consumed.
+_MIN_VALID_POSITION = 0
+_MAX_VALID_POSITION = 5000
+
+# A single-frame jump larger than this is the encoder wrapping past 0/4095
+# (a ~4096-step delta), not real motion — see CalibrationDiscontinuityError.
+_MAX_POSITION_JUMP = 2000
+
+# A recorded min..max sweep smaller than this many encoder steps means the joint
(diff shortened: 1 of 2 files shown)


nicolas-rabault (approved):
Good job.

PR #90, opened 2026-07-26
feat(onboarding): Add a guided onboarding tour for newcomers

cn0303:
## What this adds

A guided, interactive onboarding tour that walks a first-time user through the whole LeRobot loop
inside LeLab. From picking a robot, calibrating it, recording a dataset, training a policy to running it.
All in plain language, without them having to know the jargon or hunt for where each step lives.

Following up on our conversation about lowering the barrier for newcomers (students, hackathon
first-timers, anyone unboxing an SO-101), this is the first, self-contained piece: a real product
tour of the existing UI. It's built so a chat-style "collection assistant" can grow out of it later,
but that is deliberately out of scope here so this PR stays focused and reviewable.

## Demo

https://github.com/user-attachments/assets/8fecc9da-3ba5-4f5a-b453-d04772b752b9

## What the user sees

- On the first visit, a small "Welcome to LeLab" dialog offers a ~2-minute tour (or a quiet
  dismissal). The choice is remembered, so it never nags on return.
- Taking the tour, a spotlight dims the page and highlights the real control for each step;
  the robot selector, the calibrate button, the dataset card, the training button, the jobs list etc.
  with a short, jargon-free explanation beside it. The highlighted control stays clickable.
- The tour follows the user's real progress: it reads whether a robot is selected, calibrated,
  a dataset has been recorded, and a model trained, then marks steps done and gently moves on. Steps
  that need hardware are optional and always skippable, so someone browsing without an arm connected
  can still complete the whole tour.
- It can be replayed anytime from a `?` launcher in the landing top bar and a floating button on
  every page.

## How it works

- **No new dependencies.** The spotlight, cards, and dialog are hand-rolled on the Radix/shadcn
  primitives already in the tree.
- **Almost entirely frontend.** `frontend/dist` is left untouched (it is rebuilt from source by
  `build_frontend.yml` on merge). The only backend change is a small, additive `scope=local` mode on
  the existing `/datasets` endpoint (a pure filesystem scan, no Hub call), with test coverage — see
  below.
- **State-awareness stays off the Hub.** Progress is derived from endpoints the app already exposes,
  and the tour polls only cheap local ones while it's open (zero cost when closed): `/robots`
  (`is_clean`), `/datasets?scope=local`, and `/jobs`. It deliberately avoids `/jobs/hub` and the
  Hub-merging default of `/datasets`; tracked cloud jobs already surface in `/jobs`, and auth comes
  from the existing HfAuth context. The poll pauses while the tab is hidden.
- **Coexists with the Space.** The first-visit welcome is suppressed on the hosted HF Space, where the
  install modal already owns the first-run moment, so the two never stack.
- **Structure.** A declarative step model (`lib/onboardingSteps.ts`) names the real element to
  highlight via a `data-tour` attribute; an `OnboardingProvider` drives the step machine, navigation,
  and persistence; a `SpotlightOverlay` tracks the target across scroll/resize and routes; a
  `TourCard` positions the guidance with viewport-aware flipping. Accessible: focus moves to the card,
  it announces politely, and Escape exits from anywhere.

The change is organised as small, self-contained commits, each of which type-checks on its own:
engine → cross-page navigation → progress-awareness → first-visit welcome → motion/accessibility
polish, followed by review-driven fixes (keeping the progress poll and dataset check off the Hub,
scrolling below-the-fold targets into view, and a hint correction).

## Try it locally

```
lelab --dev
```

Open `http://localhost:8080`. To see the first-run welcome again after dismissing it, clear the flag
in the browser console and reload:

```
localStorage.removeItem("lelab:onboarding-v1"); location.reload()
```

## Testing

The backend change ships with `pytest` coverage (`tests/test_datasets.py`, `tests/test_server.py`):
the new local-only listing is tested for its `source="local"` tagging and, at the endpoint level, for
never reaching the Hub on `scope=local`. The full suite passes.

There is no frontend test harness in the repo, so the UI was verified with `tsc --noEmit`, ESLint,
`npm run build`, and a manual walkthrough via `lelab --dev`: first-run offer, per-step spotlight and
navigation (including scrolling below-the-fold targets into view), progress-driven auto-advance, the
skip/escape paths for hardware-free browsing, replay from both launchers, and behaviour under
resize/scroll. The UI is browser code with no OS-specific paths, so it behaves the same on Windows,
Linux, and macOS.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -20,6 +20,7 @@ import UpdateNotice from "@/components/UpdateNotice";
 import { TooltipProvider } from "@radix-ui/react-tooltip";
 import { ApiProvider } from "./contexts/ApiContext";
 import { HfAuthProvider } from "./contexts/HfAuthContext";
+import { OnboardingProvider } from "./contexts/OnboardingContext";
 
 const queryClient = new QueryClient();
 
@@ -33,24 +34,26 @@ function App() {
               <UrdfProvider>
                 <DragAndDropProvider>
                   <BrowserRouter>
-                    <SingleTabGuard>
-                      <TeleopStopNotice />
-                      <UpdateNotice />
-                      <Routes>
-                        <Route path="/" element={<Landing />} />
-                        <Route path="/teleoperation" element={<Teleoperation />} />
-                        <Route path="/recording" element={<Recording />} />
-                        <Route path="/upload" element={<Upload />} />
-                        <Route path="/training" element={<Training />} />
-                        <Route path="/training/:jobId" element={<Training />} />
-                        <Route path="/inference" element={<Inference />} />
-                        <Route path="/calibration" element={<Calibration />} />
-                        <Route path="/edit-dataset" element={<EditDataset />} />
+                    <OnboardingProvider>
+                      <SingleTabGuard>
+                        <TeleopStopNotice />
+                        <UpdateNotice />
+                        <Routes>
+                          <Route path="/" element={<Landing />} />
+                          <Route path="/teleoperation" element={<Teleoperation />} />
+                          <Route path="/recording" element={<Recording />} />
+                          <Route path="/upload" element={<Upload />} />
+                          <Route path="/training" element={<Training />} />
+                          <Route path="/training/:jobId" element={<Training />} />
+                          <Route path="/inference" element={<Inference />} />
+                          <Route path="/calibration" element={<Calibration />} />
+                          <Route path="/edit-dataset" element={<EditDataset />} />
 
-                        <Route path="*" element={<NotFound />} />
-                      </Routes>
-                    </SingleTabGuard>
-                    <Toaster />
+                          <Route path="*" element={<NotFound />} />
+                        </Routes>
+                      </SingleTabGuard>
+                      <Toaster />
+                    </OnboardingProvider>
                   </BrowserRouter>
                 </DragAndDropProvider>
               </UrdfProvider>
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -34,8 +34,8 @@ function App() {
               <UrdfProvider>
                 <DragAndDropProvider>
                   <BrowserRouter>
-                    <OnboardingProvider>
-                      <SingleTabGuard>
+                    <SingleTabGuard>
+                      <OnboardingProvider>
                         <TeleopStopNotice />
                         <UpdateNotice />
                         <Routes>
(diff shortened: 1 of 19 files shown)


frontend/src/App.tsx
```diff
@@ -33,24 +34,26 @@ function App() {
...
-                        <Route path="/" element={<Landing />} />
-                        <Route path="/teleoperation" element={<Teleoperation />} />
-                        <Route path="/recording" element={<Recording />} />
-                        <Route path="/upload" element={<Upload />} />
-                        <Route path="/training" element={<Training />} />
-                        <Route path="/training/:jobId" element={<Training />} />
-                        <Route path="/inference" element={<Inference />} />
-                        <Route path="/calibration" element={<Calibration />} />
-                        <Route path="/edit-dataset" element={<EditDataset />} />
+                    <OnboardingProvider>
```
nicolas-rabault: `OnboardingProvider` renders `WelcomeDialog`, `SpotlightOverlay` and the floating launcher as siblings of `children`, so all three sit outside `SingleTabGuard`.

On a first visit in a second tab that wedges the app. The welcome dialog mounts at `z-50`, behind the guard's `z-[9999]` blocker, so it is invisible, while Radix's modal sets `pointer-events: none` on `body` and re-enables it only on the dialog content. The guard's "Use this tab" button is outside that content, so it stops responding and the app looks frozen with no visible cause. Escape does clear it, but Escape also writes `dismissed`, so that user never gets offered the tour again.

Repro: `localStorage.removeItem("lelab:onboarding-v1")`, close all LeLab tabs, open one, then open a second. The second tab shows "LeLab is already open in another tab" and the button does nothing.

Moving `OnboardingProvider` inside `SingleTabGuard` fixes it and keeps the `?` button off a blocked tab as well.

nicolas-rabault (changes requested):
Thanks for this, and for how much restraint is in it.

One thing needs fixing before this can land, inline: the tour surfaces render outside `SingleTabGuard`. On a first visit in a second tab the welcome dialog mounts invisibly behind the guard's blocker and Radix's `pointer-events: none` kills the "Use this tab" button, so the app looks frozen. Escape clears it but also permanently writes `dismissed`. That's a newcomer locked out on first contact, by the onboarding feature, so it has to go.
Moving the provider inside the guard should do it.

Fix the tab one and I'll take another look.

frontend/src/App.tsx
cn0303: Finally done, was quite preoccupied.

I moved OnboardingProvider inside SingleTabGuard as you suggested. However, the guard still rendered children on a non-primary tab, so the welcome dialog still mounted there and Radix still set pointer-events: none on body. So I also changed the guard to render only its blocker when the tab isn't primary ({isPrimary && children}).

Now the provider, and any modal it opens, never mounts on a blocked tab: the body stays interactive, "Use this tab" works, and the ? launcher and spotlight are gone from that tab too. 

Let me know if this works better.

nicolas-rabault (approved):
Very nice, thank you @cn0303

PR #105, opened 2026-09-10
feat(datasets): browse local episodes with interrupted-recording handling

nicolas-rabault:
Continue #68 so local recordings can be browsed in-app without uploading to the Hub, including camera frames and episode navigation. Includes @ravediamond’s interrupted-recording fix (16e4644), merges current main, and refreshes the committed frontend bundle. Original commits and authors are preserved; #68 remains open while this replacement is validated.

Also keeps the episode pagination clear of the floating tour button and pins two compatible browser-data lockfile entries to versions accepted by the registry’s release-age policy.

Validation: 268 Python tests pass; all pre-commit checks pass; TypeScript and targeted episode-browser ESLint pass; production build reproduces the committed bundle. Tested the production UI against a real 50-episode dataset, including frame and episode navigation. Full frontend lint still reports six existing errors in unchanged main files.

API URL overrides and saved API URLs now accept only HTTP(S), addressing the CodeQL findings on video sources. Four focused regression cases cover rejecting JavaScript/data URLs and preserving HTTP(S) endpoints.


frontend/src/hooks/useEpisodes.ts
```diff
@@ -0,0 +1,89 @@
...
+    listEpisodes(baseUrl, fetchWithHeaders, repoId)
+      .then((r) => {
+        // The backend reports a readable dataset it couldn't parse as
+        // success:false rather than an HTTP error, so check the flag too.
+        if (!r.success) {
+          setData(null);
+          setError(r.message ?? "Could not read dataset");
+          return;
+        }
+        setData(r);
```
nicolas-rabault: [P2] Ignore obsolete dataset-list responses. If A is requested, then B, and A finishes last, setData(r) replaces B's episodes and statistics with A's while the selected dataset remains B. I reproduced this with the current hook and deferred responses. Retained data also lets EditDataset select A's first episode when switching to B. Cancel or ignore superseded requests, and only expose/use data whose repo_id matches the current selection.

frontend/src/components/dataset/EpisodeViewer.tsx
```diff
@@ -0,0 +1,401 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [detail.episode_index, activeCam, fromTs]);
+
+  const handleTimeUpdate = useCallback(() => {
+    const v = driverRef.current;
+    if (!v) return;
+    // Loop inside this episode's window rather than running on into the next
+    // episode's footage, which shares the file.
+    if (v.currentTime >= toTs) {
+      v.currentTime = fromTs;
```
nicolas-rabault: [P2] Resume playback when looping at the end of an MP4. The browser has already paused when the final episode reaches the file boundary, so setting currentTime alone rewinds without restarting. I reproduced this in the production UI on episode 49 of lerobot/svla_so101_pickplace: playback returned to 388.933333s and stayed paused. Handle the ended boundary so an episode that was playing continues looping, including single-episode files, and add a regression check.

PR #106, opened 2026-09-11
feat(record): settle before the first episode, and clarify the recording screen

cetiennec:
## What

Recording jumped straight from *preparing* into writing frames, so the first seconds of episode 1 caught the arm still being moved into position.

Every later episode already gets that grace period: the **reset phase** between episodes runs the control loop with no dataset, so the arm keeps following the leader while nothing is recorded. The first episode simply never had one. It does now — same phase, same `reset_time_s`, same orange in the UI, same exit-early control to skip it.

No new setting on purpose: a second duration for what is mechanically the same phase would only mean two orange stretches back to back between episodes.

## Cleanup

The loop called `record_loop` in four places with the same eight arguments each time, differing only in `dataset` and `control_time_s`. `loop_kwargs` collects the shared part once; each call site keeps the real `record_loop(**loop_kwargs, ...)` call and states just what differs. (Deliberately not `functools.partial`: that would leave exactly one `record_loop(` in the file and make the four real call sites ungreppable by name.)

## Recording screen

- **Episode counter** is a boxed chip instead of small gray text
- **Phase limit** reads at a glance next to the elapsed time (`00:12 / 00:30`), with a caption naming what is being timed, so it isn't confused with the session clock
- **Re-record** is a real button next to the primary action (**DEL** or **←**) rather than a row in the overflow menu
- The startup screen says which half of the handshake is stuck, instead of spinning *Connecting...* with no way back

## Fix (review feedback)

Both "Back to Home" exits used to navigate away unconditionally, leaving an active session's worker thread driving the arm with no UI attached to it:
- **Startup screen**: if `/start-recording` had already succeeded — or succeeded *after* the click while still in flight — leaving abandoned it.
- **Main screen**: once a session is confirmed live, leaving now fires `/stop-recording` first.

Two follow-up fixes from a second review pass on that:
- `/start-recording` and `/stop-recording` both always answer HTTP 200, even on failure — so `response.ok` couldn't tell success from failure. Neither path was checking it before treating a stop as done; an HTTP error status now routes through the same catch-driven retry as a network failure.
- A rejected `/start-recording` fetch doesn't prove the server never created a session (the request may have reached it before the *response* was lost). That case now stays "uncertain" and is treated as if a session might exist — routed through the stop-then-leave path — rather than being read as safe to skip.

## Notes

- `frontend/dist/` rebuilt with Node 22, as the Quality workflow requires.
- Verified end to end against a simulated SO-101 pair (settle-in → record → re-record → reset → next episode); the simulation harness itself is not part of this PR.


frontend/src/pages/Recording.tsx
```diff
@@ -476,13 +492,28 @@ const Recording = () => {
...
+          <p className="text-lg">
+            {recordingSessionStarted
+              ? "Connecting to recording session..."
+              : "Starting recording session..."}
+          </p>
+          {startupError && (
+            <p className="mt-4 text-sm text-red-400">{startupError}</p>
+          )}
+          <Button
+            onClick={() => navigate("/")}
```
nicolas-rabault, line 511: [P2] Stop the session before leaving the startup screen. If /start-recording succeeds but /recording-status fails, this button navigates home without sending /stop-recording. Unmounting only cancels polling, so the recording worker keeps driving the arm and cycling through takes without its recording controls. The same race exists if the start request is still pending when the button is clicked. Make this exit cancel the session, including a late successful start response, and retain a way to stop it if cancellation fails.

nicolas-rabault (changes requested):
Thanks for the contribution! The recording transition checks passed. Please address the startup exit issue before merging; details inline.

nicolas-rabault (changes requested):
Thanks for the contribution! The recording transition checks passed. Please address the startup exit issue before merging; details inline.

frontend/src/pages/Recording.tsx
cetiennec: This should be fixed by the last commit, testing this afternoon

cetiennec:
@nicolas-rabault tested on a real robot and adressed the comment

frontend/src/pages/Recording.tsx
```diff
@@ -288,6 +322,58 @@ const Recording = () => {
...
+    }
+
+    try {
+      const response = await fetchWithHeaders(`${baseUrl}/stop-recording`, {
+        method: "POST",
+      });
+      // This endpoint always answers 200; `success: false` in the body just
+      // means there was nothing left to stop (e.g. a retry after it already
+      // did) — either way it's now safe to leave.
+      await response.json().catch(() => null);
```
nicolas-rabault, line 364: [P2] Keep the stop control available after an HTTP error. fetch resolves normally for HTTP 500/502 responses, so this path marks the session inactive and navigates home even when the stop request failed. handleBackToHome has the same issue at lines 548-549. Check response.ok in both paths and route unsuccessful responses through the existing error/retry behavior before navigating.

frontend/src/pages/Recording.tsx
```diff
@@ -312,6 +413,11 @@ const Recording = () => {
         navigate("/");
       }
     } catch (error) {
+      activeSessionRef.current = false;
```
nicolas-rabault: [P2] Preserve uncertainty when the start response is lost. A rejected fetch or unreadable response does not prove the server failed to start recording. If the user clicks Back to Home while start is pending and the response then drops after the server accepted it, setting this ref to false makes attemptLeave navigate home without ever sending stop. Keep this state uncertain and retain cancellation/retry until the backend confirms the session is stopped or absent.

nicolas-rabault (changes requested):
Thanks for following up! Two failure paths still need attention; details inline.

cetiennec:
Can you recheck @nicolas-rabault ?

PR #107, opened 2026-09-11
added camera FPS settings to recording and inference instead of hardcoded 30fps, naming

bellabzhu:


bellabzhu:
Sorry, this pull request was a mistake. If you can delete it, I would appreciate.

